PR #6, opened 2017-09-14
Add pagination

mikew:
I noticed the `all` functions wouldn't actually list all of a resource.

### DIFF
--- a/src/API.js
+++ b/src/API.js
@@ -1,5 +1,6 @@
 const Request = require('request-promise');
 const { Groups, Projects, Issues, Runners, Users, Labels } = require('./Models');
+const { defaultPaging } = require('./Utils');
 
 function defaultRequest(url, endpoint, { headers, body, qs, formData }) {
   const params = {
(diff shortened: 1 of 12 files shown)


jdalrymple:
Lol I was just implementing this!

jdalrymple:
I'll look over this in the next couple days, looks good though :)

jdalrymple:
Just pushed my implementation, noticed it can cause problems  if there are many results.  maybe i should add some timeout option?

mikew:
Hmm yeah, a timeout would help. Would another option for max number of pages make sense?

jdalrymple:
What would be the use case of max number of pages? Wouldn't you want to leave that up to the user of the library? If they only wanted 2 pages, wouldn't they just pass in the per_page and page options? Hmm, I guess it would become tedious. 

Im not sure whats the best way to handle both functionalities. :confused:

jdalrymple:
What about no time out, just include limit ( which limits the amount of results not pages)

jdalrymple:
Alright, for now ill just add a limit argument, which will limit the amount of total results. If it still causes problems we could always add a timeout of some sort

jdalrymple:
Added pagination in the latest release, tell me if it works for you !

PR #9, opened 2017-11-12
Fix typo in project sharing function

christophlehmann:
See https://docs.gitlab.com/ce/api/projects.html#share-project-with-group for possible options


jdalrymple:
Nice catch!!

PR #10, opened 2017-11-16
Clarifications and additional examples to the readme

FrankV01:
Attempting to provide alt. examples and some clarification on the token for others'. The provide examples didn't work for me (with `await`) but using the promise notation directly helped. Also helps show others' that it's a promise that is returned.


jdalrymple:
Hmm, yea it's misleading. Await is not available outside of  function . I'll add more clarification to the readme!

Issue #19, opened 2017-12-02
Options missing from listCommits(...)


sparcs360:
Hi

Can you add an `options` parameter to `ProjectRepository.listCommits` so I can do some date filtering please (see https://docs.gitlab.com/ee/api/commits.html)?

```javascript
  listCommits(projectId, options = {}) {
    const pId = Utils.parse(projectId);

    return this.get(`projects/${pId}/repository/commits`, options);
  }
```

Then it can be called with:

```javascript
    const options = {
        "ref_name": "...",
        "since": "..."
    };
    return gitLab.projects.repository.listCommits(gitlabProject, options).then(...)
```

Much appreciated!

jdalrymple:
Will do! Will be in the 2.0.x release :)

jdalrymple:
Just released!

PR #21, opened 2018-01-03
Add options for getting commits

MartinBenninger:
Add the options parameter to requests getting all commits and all commit comments.

This is needed in order to set the paging options (and others).

I believe this functionality was intended to be implemented as mentioned in the comment on issue #19.


jdalrymple:
Looks good!

PR #22, opened 2018-01-04
Remove unneeded console.log call

MartinBenninger:
I noticed this function was being logged to the console and I assume it's just left over from testing.


jdalrymple:
Nice catch! Thank you

PR #23, opened 2018-01-10
Added project-repository-files docs page

arsdehnel:
Hi @jdalrymple I like this project over some other gitlab api projects and like that you're trying to document things within the project rather than assuming people can figure out the details of how to make the calls by reading the code.  

As I documented this call that I'm making I can't help but wonder if the `encodeURIComponent` should be part of this package rather than having users have to wrap their path in that.  I'd be happy to make that change but wanted to run it by you first.  Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,3 +312,33 @@ inferred from the file path
 - Initial release
 - TODO: Tests, Examples
 
+## Development
+
+To get this running locally rather than from your `node_modules` folder:
+
+```bash
+$ git clone https://github.com/jdalrymple/node-gitlab-api.git
+$ cd node-gitlab-api
+$ npm install
+$ npm build
+```
+
+And then inside whatever project you are using `node-gitlab-api` in you change your references to use that repo.  In your package.json of that upstream project change:
+
+```json
+  "dependencies": {
+    ...
+    "node-gitlab-api": "2.1.0"
+    ...
+  }
+```
+
+to this
+
+```json
+  "dependencies": {
+    ...
+    "node-gitlab-api": "<path-to-your-clone>"
+    ...
+  }
+```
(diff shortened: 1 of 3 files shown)


jdalrymple:
Thank you! It was one of the main problems I ran into when I was trying to use the GL API. I figured adding a few extra docs couldn't hurt lol. I think it should be part of the package for sure! It would definitely help those using the library by removing the annoying step of making their paths encoded.

arsdehnel:
Cool.  I can add that change to this PR and then also update this docs section to reflect that change.  might also see about adding more to this page so it's not just the lonely one call on there.

jdalrymple:
Sounds good! Thanks again!!

arsdehnel:
hey @jdalrymple wondering if you have recommendations on a setup for development?  I'm still quite the noob on working on npm packages locally.

arsdehnel:
got a working setup and documented it in the readme.  not sure the current best practice but looks like either [npm link](https://docs.npmjs.com/cli/link) or [linklocal](https://www.npmjs.com/package/linklocal) are the leading contenders.  If you have suggestions on the best setup I'll do that and then document what I've got.

I also updated the code and my docs contribution, feedback welcome on those changes.

jdalrymple:
Ill check through your work tonight. At a glace it looks great!! As for the linking, it would probably be best to stick with the npm script as a standard, so npm link.

arsdehnel:
@jdalrymple these are indeed pretty similar lines but this pattern is used throughout the project so I'd hate to do something new here.  Thoughts on changing how I did this vs following the pattern of the project?

jdalrymple:
Still haven't had a chance to read through your PR yet due to some personal commitments :(, will do soon though!!

jdalrymple:
Looks good! Ill try and clean up the code climate stuff!

PR #41, opened 2018-02-08
Added delete key method to UserKeys.js

claude-abounegm:
Just a small edit to `UserKeys.js` which allows to delete an SSH key from a User.

# Usage
`await gitlabApi.users.keys.remove(userId, userKey);`

# Gitlab Docs
https://docs.gitlab.com/ce/api/users.html#delete-ssh-key-for-given-user

### DIFF
--- a/src/Models/UserKeys.js
+++ b/src/Models/UserKeys.js
@@ -11,6 +11,13 @@ class UserKeys extends BaseModel {
     });
   }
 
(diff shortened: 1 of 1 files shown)


jdalrymple:
Woo thanks!! I'll release later this week !

claude-abounegm:
Awesome, thanks!

PR #45, opened 2018-03-19
Added optional parameters for adding and editing resource members

Marethyu1:
I noticed that in the [resource members](https://github.com/jdalrymple/node-gitlab-api/blob/master/src/Models/ResourceMembers.js) class there is no support for optional params. More specifically, you can't add the [expires_at](https://docs.gitlab.com/ee/api/members.html#add-a-member-to-a-group-or-project) property to a request. Thought this could be a simple solution for optional parameters, which could be reused in other places. Happy to update documentation to reflect these changes if you like it!

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -8,6 +8,7 @@
   ],
   "plugins": [
     "@babel/plugin-proposal-export-default-from",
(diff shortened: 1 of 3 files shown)


jdalrymple:
Good catch, thanks!

PR #47, opened 2018-03-21
fix(Doc): correct List all group projects api call

niklas-dahl:
the current code gives `err Cannot read property 'projects' of undefined`


jdalrymple:
Oops!! Thanks

PR #50, opened 2018-03-22
Fix current user "keys" routes

Mr-Wallet:
As per the [docs](https://docs.gitlab.com/ee/api/keys.html), all user key routes have as a substring either `users/:id/keys` (plural `users` with `:id`) or `user/keys`(singular `user` with no user ID).

This PR fixes a typo where the latter form had `users` pluralized, which resulted in a 404 from Gitlab.com.


jdalrymple:
Thanks! Not sure how I missed that 🙃

PR #52, opened 2018-03-24
3.0.0

jdalrymple:
Major release. See Readme for changes


jdalrymple:
Noticed a conflict between the Namespace API and the Namespace exports. Will come up with another name.

jdalrymple:
Replaced the name Namespace with Bundle for now

jdalrymple:
Suggestion: Using collection instead of bundle?

jdalrymple:
Removed theold docs since they were quite outdated. Added a note in the readme with the plan to flush out these docs as the next big project for this library

PR #54, opened 2018-03-27
Fix getting forks of a project

Mr-Wallet:


jdalrymple:
Will merge and publish tonight

PR #59, opened 2018-04-02
Add XMLHttpRequest support

### LINKED ISSUE #56: XMLHttpRequest support
This is great for electron apps (or basically any situation where you know the code is running in a browser, because you can exploit the browser's proxy resolution and SSL certificates. It's not normally available in Node, however, so it would have to be an option.

We swapped to this package and a few of our users are complaining about failures with self-signed certificates.

Basically this would be a constructor option that switches out the core driver of the requests. We did this on the original repository with a mix of PRs and fork-fixes for its `node-slumber` dependency. I totally expect to write this myself but I wanted to open the issue in advance to make sure you're amenable to such an extension to this package.

Mr-Wallet:
Fixes #56

This is a work in progress - I haven't even run it to confirm that it works. However I wanted to get input on this as early as possible so I can make any adjustments as needed.

It was a little hard to figure out where things should go, since RequestHelper tries to be "pure" and the service is "stateful", but I need the fundamentals of how to make requests to be somewhat stateful.

Eventually I went with abstracting the "requester" to the constructed service. The most popular XHR package and Request _mostly_ overlap in the ways this package is using Request, except that `xhr` does not return promises, and it has no support for `qs` (it may also not have support for `formData`, but we have no need for that and we can cross that bridge later).

The `qs` issue is resolved on `xhr`'s development branch for version 3, but the current stable version is 2, so for now we need some way to switch how the querystring gets applied. Since `xhr` is moving in the direction where we won't need to do anything special in the future, I didn't want to over-engineer it; I simply pass through the `useXMLHttpRequest` flag through to `defaultRequest` via the `service`, so I can do some manual modification. It's not a great pattern, but this way, in a future version after `xhr` gets updated, all of that logic can simply be deleted without any more rewrite.

The packages haven't been updated yet to reflect the new `xhr` and `qs` dependencies, simply because I haven't gotten someone on a Mac to commit it yet (on my Windows machine, the `package.lock` loses `chokidar/fsevents`.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ const Gitlab = require('node-gitlab-api/dist/es5').default
 // Instantiating
 const api = new Gitlab({
   url:   'http://example.com', // Defaults to http://gitlab.com
-  token: 'abcdefghij123456'	//Can be created in your profile. 
+  token: 'abcdefghij123456'	// Can be created in your profile.
 })
 
 // Or, use a OAuth token instead!
@@ -160,7 +160,7 @@ import { Projects } from 'node-gitlab-api';
 
 const service = new Projects({
   url:   'http://example.com', // Defaults to http://gitlab.com
-  token: 'abcdefghij123456' //Can be created in your profile. 
+  token: 'abcdefghij123456' // Can be created in your profile.
 })
 
 ```
@@ -175,7 +175,7 @@ import { ProjectsBundle } from 'node-gitlab-api';
 
 const services = new ProjectsBundle({
   url:   'http://example.com', // Defaults to http://gitlab.com
-  token: 'abcdefghij123456' //Can be created in your profile. 
+  token: 'abcdefghij123456' // Can be created in your profile.
 })
 
 services.Projects.all()
@@ -253,6 +253,22 @@ EpicNotes
 EpicDiscussions
 ```
 
+### Using XMLHttpRequest
+This package uses the [Request](https://github.com/request/request) library by default, which is built into Node. However, if your code is running in a browser, you can get better built-in resolution of proxies and self-signed certificates by using the browser's XMLHttpRequest implementation instead:
+
+```javascript
+import Gitlab from 'node-gitlab-api';
+
+const api = new Gitlab({
+  url:   'http://example.com', // Defaults to http://gitlab.com
+  token: 'abcdefghij123456',	// Can be created in your profile.
+
+  useXMLHttpRequest: true // Use the browser's XMLHttpRequest instead of Node's Request library
+})
+```
+
+**WARNING:** Currently this option does not support the `multipart/form-data` content type, and therefore the endpoint for [uploading a file to a project](https://docs.gitlab.com/ee/api/projects.html#upload-a-file) will not work correctly. All other endpoints should work exactly as expected.
+
 ### Examples
 Once you have your library instantiated, you can utilize many of the API's functionality:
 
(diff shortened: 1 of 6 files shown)


jdalrymple:
This is a smart way to get the functionality and maintain flexibility for future improvements!

Moving it into its own constructor would be cleaner but that can be easily added latter as a refactoring without changing much as well. 

Looks good !

jdalrymple:
Will need to add some info in the docs about this, especially in reference to the functionality that doesn't work for the time being with the XMLhttprequest library

Mr-Wallet:
From the end user's perspective I believe it should be totally seamless, except I'm not so sure about `formData`. We're not putting that into any of our requests and honestly I don't know what in the API it's used for. It's not clear why anyone would need, in this library, to specifically send things as `application/x-www-form-urlencoded` or `multipart/form-data`, which seems to be the effect of defining `formData` in Request.

At any rate, there's nothing about `formData` in the docs of xhr so I don't think the library will respect that option. Besides that and mentioning the XHR constructor flag I don't think any README changes will be necessary.

I'm still wrestling with self-signed certs but I'm hoping to get this in shape tomorrow (Tuesday).

jdalrymple:
lol formData is used for only one endpoint, the upload endpoint on projects.

Mr-Wallet:
Ah, that makes sense. Checked the gitlab docs and xhr will probably not work with that endpoint without more work. If it's all right with you, for now I'll just note that deficiency in the README where the xhr option is documented.

jdalrymple:
Yea thats fine, as long as its documented I have no issue with it. Something for a future pr

Mr-Wallet:
Well, uh... glad you're not hyper-optimizing package size. 😅 

I think this should be good to go. Finally got our IT issues sorted out and confirmed that using XHR with this change resolves our issue.

jdalrymple:
😂 can always optimize later

jdalrymple:
Will review it tonight !

jdalrymple:
Left a few comments, but otherwise looks  good!!

jdalrymple:
Closed by accident!

Mr-Wallet:
It looks like our `delete` code broke at some point between v2 and here. I believe the problem is that `RequestHelper` is returning just the `body`, which is similar to the old package by moul, but the other methods are returning the entire response.

I think the fix is to make the other methods besides `get` also await the response and then return just the body.

Unfortunately, fully half of that solution is in conflict with this PR, so I was hoping either this can get merged, or you would be willing to let me tack that fix onto this PR even though it's not related to XHR.

jdalrymple:
I can create a separate PR to fix the problem. Something is definitely off since all the functions should only be returning the body. The flag to return the full response is defaulted to false and is only every overridden with the get function when doing pagination. 

https://github.com/jdalrymple/node-gitlab-api/blob/master/src/infrastructure/RequestHelper.js#L9

Mr-Wallet:
D'oh, I bet the issue is that my manually promisified `xhr` doesn't care about that option. IIRC that's not even part of `request`, it's part of `request-promise`.

So the solution probably _does_ belong in this PR - I need to make `XMLHttpeRequester` respect that flag and behave accordingly.

jdalrymple:
I would love to use just request, but it doesnt support promises *sigh*

Mr-Wallet:
I haven't tested that the above commit works, but that's my sketch for how I think to resolve the problem.

In experimenting with this problem I found that "current user keys" is broken again. [aeebf6](https://github.com/jdalrymple/node-gitlab-api/commit/aeebf67f571f5fd803e0c06dc9bf78c860396a14) swapped the URL back to pluralized `users/keys`.

jdalrymple:
Ill fix it now and release! The latest release had so many changes :disappointed:

Mr-Wallet:
I think this is good to go... we have these changes privately on top of `rc7` and they seem to be working.

jdalrymple:
Alrighty, ill merge it in tonight and see whats up :)

jdalrymple:
Merged and published!

PR #64, opened 2018-04-10
first crack at some testing

arsdehnel:
This is by no means intended to be merged.  This was just meant to get a starting point out there and see what you think.  Added some brief words on the readme about how to fire it up for testing.  I'm using `jest` to run the tests which is what we use at work all the time so it's just what I'm familiar with. Otherwise this is a super basic test just proving out that it can run and hit the dockerized gitlab.

### DIFF
--- a/README.md
+++ b/README.md
@@ -664,3 +664,24 @@ to this
     ...
   }
 ```
+
+## Testing
+
+Testing is a work-in-progress right now but here is the start.  
+
+1. First run gitlab in a docker container:
+
+```bash
+docker-compose -f docker-compose.test.yml up
+```
+
+1. Once GitLab is up on localhost:8080, create the admin password, then login as `root` with that password and then create an access token ( Avatar -> Settings -> Access Tokens ).  
+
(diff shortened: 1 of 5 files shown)


jdalrymple:
This is really cool! Reading up on jest now :o

jdalrymple:
Should device a way to preconfigure the gitlab build to already have a token available

arsdehnel:
Yup, agreed. Wanted to make sure the rest of the approach seems good before spending too much time on it.

jdalrymple:
The only thing I'd change is the folder structure, and maybe switching to a tap reporter because I for some reason really like tap 🤣 but besides that looks really good. Would travis ci then be responsible for spinning up the gitlab instance through docker compose  and then running the library against it.  Or would the getting the library into the container also be part of a docker compose file ?

jdalrymple:
I added more tests to the gitlabci branch! :smile:

jdalrymple:
Weird, tried to merge my changes into your remote branch and that autoclosed the PR. hmmm

PR #71, opened 2018-04-19
Beginning development of Testing framework

jdalrymple:
Continuation of PR #64


jdalrymple:
Still having an issue with babel 7+ and jest :(

jdalrymple:
```
Requires Babel "^7.0.0-0", but was loaded with "6.26.3". If you are sure you havea compatible version of @babel/core, it is likely that something in your build process is loading the wrong version. Inspect the stack trace of this error to look for the first entry that doesn't mention "@babel/core" or "babel-core" to see what is calling Babel.
```

jdalrymple:
Figured it out!! Just debugging the tests now

jdalrymple:
Added the first set of tests. More will be added in the future now that the system is setup :)

PR #76, opened 2018-04-24
Add some Jobs related endpoint support

yonguelink:
We needed a few endpoint support related to Jobs, supposed we should add them to the project.

Add support for :

1. Listing the job from a specific pipeline via `PipelineJobs.all` <- The `PipelineJobs` class is discutable, we weren't sure where we should've placed the "get all jobs from a pipeline" endpoint...
1. Download the content from a specific file from the artifacts of a specific job via `Jobs.downloadSingleArtifactFile` 
    * Also added a way to download this single file using streams via `Jobs.downloadSingleArtifactFileStream`
1. Playing a specific job via `Jobs.play`

I tested both manually, they're working.

References:

* https://docs.gitlab.com/ee/api/jobs.html#list-pipeline-jobs
* https://docs.gitlab.com/ee/api/jobs.html#download-a-single-artifact-file
* https://docs.gitlab.com/ee/api/jobs.html#play-a-job


yonguelink:
We might need to add more endpoint support in a near future. Perhaps it is better not to merge is yet... I kind of clicked on submit PR a tad too quickly 😅

jdalrymple:
lol, If there is only one pipelinejob endpoint, may be easier to add it to the jobs api as well

yonguelink:
I went over the doc at gitlab, and yeah it seems they did group it like you suggested as there was only one request for that api... 

I would have a tendency to place that under the `Pipelines` api tho, something like `Pipelines.showJobs`. I think this is more logical than `Jobs.showPipelineJobs`. What do you think?

jdalrymple:
Yup that makes sense too! Could put it under both if you think it would be helpful as well !

⁣Sent from TypeApp ​

yonguelink:
I went ahead and put it at both places.

I also went and did a 'streamGet' to be able to download the artifacts and place them directly in a file, preventing memory usage for the whole artifacts (imagine having a few hundred MBs to download and keep in memory! If not GBs). 

While doing that `streamGet` point, I wasn't sure what was going on in the normal `get` and now I'm uncertain the stream endpoint does all the job it needs to do. It works for our use-case (`streamGet.pipe(fileStream)`) tho.

yonguelink:
Hm. I've been, finally, added to my organization' github. I'm gonna close this PR and we'll create a new one once we're done adding stuff. Sorry for bothering you! 😄

Issue #89, opened 2018-05-18
Migrate to TypeScript


r00ger:
Track progress/ideas/questions under this ticket

MunifTanjim:
May I ask what will be the primary benefit of moving the whole codebase to TS?

jdalrymple:
I believe the initial goal was to improve documentation efforts

MunifTanjim:
If we automate the routes generation process, converting the whole codebase to TS won't have much benefit in terms of "improving documentation efforts".

But if this is being done for any other benefits then I've got nothing to say other than, "_TS is really cool_" :smiley:

jdalrymple:
:joy: fair. Either way, trying to focus on tackling one thing at a time.

r00ger:
Hey! Do you have a plan to merge `next` branch to trunk any time soon?

jdalrymple:
Yup! Just waiting until I have the time to test through it all before I release it

r00ger:
@jdalrymple will you please rebase `next` on the latest `master`?

jdalrymple:
Will do today!

jdalrymple:
Going to close this since there is a dup case

PR #90, opened 2018-05-18
Remove non-standard Babel's plugins

r00ger:
In order to migrate to TypeScript, we have to get rid of non-standard language features

#89


r00ger:
Typescript support stage3+ proposals

r00ger:
They don't have plugins, features are being implemented in compiler itself, that's why they pick only standard features

jdalrymple:
Ah, makes sense! Do you foresee that as a future problem? Where typescript lags behind such plugins? I dont know enough about the dev cycles of typescript :)

r00ger:
About feature support — it neven was an issue to me (for more than two years). Frankly, I don't care about proposal (as the two I'm deleting), but when it comes to rather important features as spead/async/await/generators the lag isn't a problem, they release every two months

PR #96, opened 2018-05-27
Typescript

r00ger:
#89


jdalrymple:
Did you drop the lodash dependency for any reason in particular ?

max-wittig:
What's the reason for this re-write? More dependencies and transpilers will be required :/

r00ger:
@max-wittig main reason is self-documenting. Also, ease of development and type-safety.

We'll have fewer dependencies. `tsc` is self-sufficient, and once we're done, Babel with its' plugins is gone

jdalrymple:
@r00ger I see, would be nice to write some helper function however. The implicit declaration is tedious lol

r00ger:
Sure, it'd be much simpler to reimplement it and annotate properly. We also can hook into lodash's typings but it's pointless in this case

r00ger:
@jdalrymple please take a look

jdalrymple:
Will do tonight!!

One question thus far (I think i asked it somewhere else, but I cant seem to find it lol) You modified the references to the spread operator (...variableName) citing that it needs to work with TS, but doesnt TS handle the transpiling of es2017 anyways? lol My knowledge of TS is still a work in progress

r00ger:
For the moment, there is (almost) no types added to make it simpler to review

[`declarations` flag](https://www.typescriptlang.org/docs/handbook/compiler-options.html) set to false, so that we don't break typings (yet) for these who use [@types/gitlab](https://www.npmjs.com/package/@types/gitlab)

There is tsconfig.json, which sets strict to true, thus showing type errors in IDE. Also, there are two other configs which turnes off almost all important type chechs so that we can compile not typed code.

I had to remove/replace some modern language features as there are some differences in `tsc` and `babel`.

For example, `Object.assign({}, foo) → {...foo}`, as Typescript transpiles spread operator to downlevel js, but do not automatically transpile ES6 methods, showing you a compilation error. This means, we'll have to polyfill ES6 methods that we want to use and then update typings, so that there is no compilation errors.

Also `[].includes → [].indexOf` for the same reason (to be fair, babel in current configuration was not replacing `includes` also, even though it's not a part of ES5).

_I will be able to answer your questions and make changes today and tomorrow, then I'll be off for a week or so_

jdalrymple:
Merged into 3.5 release. Going to release this as a rc to catch any issues since so much of the code was updated :)

PR #111, opened 2018-06-26
correct use of babel --config-file

bodtx:
There should not be a .babelrc in the root dir to enable this feature

store separate babelrc in directories

refs #67


jdalrymple:
Do you know where they specify that this is the way to do it ? Just out of curiosity?

bodtx:
No I was just trying some random tests and figure out that the default  .babelrc disbale the feature

jdalrymple:
It's just weird that the behaviour changed all of a sudden hmm

bodtx:
indeed, I've asked it [here ](https://github.com/babel/babel/pull/6133#issuecomment-400703738)

loganfsmyth:
> It's just weird that the behaviour changed all of a sudden hmm

There was a change to the way `--config-file` is implemented in Babel's beta, which you appear to have hit here, likely when you upgraded to a new beta.

The core of the issue is that `--config-file` does not disable Babel's standard logic for searching for and loading `.babelrc` files relative to the file being compiled. When you do

```
--config-file .babelrc
```
Babel will essentially load the `.babelrc` once as a config file, and once using the normal auto-searching/loading logic.

So when you do
```
--config-file .babelrc-es5
```
now you're loading `.babelrc` and _also_ `.babelrc-es5`.

What changed recently is the order in which those two are processed. Where before `--config-file` would override the `.babelrc` files, now `.babelrc` files take priority over `--config-file`.

There are a number of ways to approach fixing this. This PR is one way, though it's not what I'd recommend. My thoughts:

* Just don't name the config files `.babelrc`. If you give them any other name, they won't be auto-loaded, and they'll just be loaded as a config file.
* Leave it all in one config file and toggle environments:

```
{
  "presets": [
    ["@babel/preset-env", {
      "targets": {
        "node": "8.9.0"
      }
    }]
  ],
  "plugins": [
    "@babel/plugin-proposal-object-rest-spread"
  ],
  "env": {
    "es5": {
      "presets": [
        ["@babel/preset-env", {
          "useBuiltIns": false, 
        }]
      ],
      "plugins": [
        "@babel/plugin-transform-async-to-generator",
        "@babel/plugin-transform-runtime",
      ]
    }
  }
}
```
Then 
```
    "build:es6": "babel src -d dist/latest --config-file=./.babelrc",
    "build:es5": "babel src -d dist/es5  --config-file=./.babelrc-es5",
```
would be
 
```
    "build:es6": "babel src -d dist/latest",
    "build:es5": "babel src -d dist/es5 --env-name es5",
```

jdalrymple:
That solution is much cleaner, even more so than the one that was originally implemented! @bodtx I can implement that momentarily!

jdalrymple:
Perfect, just published, @bodtx give it a look :)

bodtx:
Great I've tested with this [project](https://github.com/jnovick/gitlab-bug-demo)
changing dependencie to `import Gitlab from "gitlab/dist/es5";`
`npm run build` is working now :+1: 

The solution is very nice, I've learned `-env-name` and `--no-babelrc` :1st_place_medal: 

Will you release ?

jdalrymple:
Already released :) gitlab 3.4.4

bodtx:
Great, will import it at work tomorrow thx

PR #115, opened 2018-07-02
Update README.md badges

MunifTanjim:
Some of the badges were for other projects (`npm`, `express`) and other tiny changes.


jdalrymple:
Thanks!!

PR #117, opened 2018-07-04
feat: obey the rate limit

max-wittig:
This PR enables node-gitlab to wait if the GitLab rate-limit is hit.
It waits as long as GitLab specifies via the [Retry-After header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After)

GitLab lacks documentation about this feature, but rate-limits can be set in the Admin UI (see screenshot)
and via the settings API.
image

```json
{
 "throttle_authenticated_api_enabled": false,
 "throttle_authenticated_api_requests_per_period": 2,
 "throttle_authenticated_api_period_in_seconds": 10,
}
```

### DIFF
--- a/src/infrastructure/RequestHelper.js
+++ b/src/infrastructure/RequestHelper.js
@@ -54,7 +54,11 @@ function getStream(service, endpoint, options = {}) {
   return StreamableRequest.get(requestOptions);
 }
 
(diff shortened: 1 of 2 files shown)


jdalrymple:
Thanks for getting around to this, been swamped with work lately and haven't had the chance to do much :(

PR #119, opened 2018-07-05
Fix methods for editing MR approval/approver settings

maclennann:
The code was previously trying to do GETs to the routes, which was giving me
404s. Updated to correctly change the `/approvals` router to POST and the
`/approvers` router to PUT as in the GitLab API documentation:

https://docs.gitlab.com/ee/api/merge_request_approvals.html


jdalrymple:
Nice catch !!

jdalrymple:
Ill release properly tonight!

PR #134, opened 2018-07-26
I needed the functionality to update push rules so I added it to Projects.

jennparise:
I needed the functionality to update push rules so I added it to Projects. It is a small change that I tested out.

### DIFF
--- a/src/services/Branches.js
+++ b/src/services/Branches.js
@@ -16,9 +16,8 @@ class Branches extends BaseService {
     });
   }
 
-  protect(projectId, branchName, options) {
-    
-    const [pId, bName] = [projectId, branchName].map(encodeURIComponent);
+  protect(projectId, options) {
+    const pId = encodeURIComponent(projectId);
     
     return RequestHelper.post(this, `projects/${pId}/protected_branches`, options);
   }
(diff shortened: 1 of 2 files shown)


jennparise:
I also updated protect branches as the previous way did not work for me. It matches the current documentation for the api now.

jdalrymple:
Woo perfect! I'll review it in the next day or so! Thank you!!

src/services/Branches.js
```diff
@@ -16,14 +16,10 @@ class Branches extends BaseService {
     });
   }
 
-  protect(projectId, branchName, options) {
-    const [pId, bName] = [projectId, branchName].map(encodeURIComponent);
+  protect(projectId, options) {
+    const pId = encodeURIComponent(projectId);
```
jdalrymple: Based on this: https://docs.gitlab.com/ee/api/protected_branches.html#protect-repository-branches, the name of the branch is still mandatory. So instead of just having the generic options could it be more like:

```
 protect(projectId, branchName, options) {

     return RequestHelper.post(this, `projects/${pId}/protected_branches`, {
         name: branchName, 
         ...options
      });
}
```

jennparise: sure, I can update it.

PR #143, opened 2018-08-09
Add Push Rule service

casz:
Love the newly added `updatePushRule` but it felt lacking when you wanted to enforce push rule for all projects and gitlab API was lacking a `createOrEdit` function and would return null with status code 404 when no push rules were created.

### DIFF
--- a/src/services/PushRule.js
+++ b/src/services/PushRule.js
@@ -1,0 +1,37 @@
+import { BaseService, RequestHelper } from '../infrastructure';
+
+class PushRule extends BaseService {
+
+  create(projectId, options) {
+    const pId = encodeURIComponent(projectId);
+
+    return RequestHelper.post(this, `projects/${pId}/push_rule`, options);
+  }
+
+  async edit(projectId, { upsert, ...options } = {}) {
+    const pId = encodeURIComponent(projectId);
+
+    if (upsert) {
+      const pushRule = await this.show(projectId);
+
+      if (!pushRule) return this.create(projectId, options)
+    }
+
+    return RequestHelper.put(this, `projects/${pId}/push_rule`, args);
+  }
+
+  remove(projectId) {
+    const pId = encodeURIComponent(projectId);
+
+    return RequestHelper.delete(this, `projects/${pId}/push_rule`);
+  }
+
+  show(projectId) {
+    const pId = encodeURIComponent(projectId);
+
+    return RequestHelper.get(this, `projects/${pId}/push_rule`);
+  }
+
+}
+
+export default PushRule;
--- a/src/services/index.js
+++ b/src/services/index.js
@@ -63,6 +63,7 @@ export { default as Services } from './Services';
 export { default as Tags } from './Tags';
 export { default as Todos } from './Todos';
 export { default as Triggers } from './Triggers';
+export { default as PushRule } from './PushRule';
 
 // General
 export { default as ApplicationSettings } from './ApplicationSettings';
--- a/test/tests/services/PushRule.js
+++ b/test/tests/services/PushRule.js
@@ -1,0 +1,18 @@
+import { PushRule } from '../../../src';
+
+describe('PushRule', () => {
+  it('should create or edit push rule on upsert', async () => {
+    const service = new PushRule({
+      url: process.env.GITLAB_URL,
+      token: process.env.PERSONAL_ACCESS_TOKEN,
+    });
+    const pushRules = {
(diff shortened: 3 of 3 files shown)


casz:
did a npm test with the newly added createOrEdit test and works on our instance. Guess the Travis CI is having a bad day 😢

jdalrymple:
Yea travis is still having issues and im not sure why. Still waiting to hear back from them

src/services/PushRule.js
```diff
@@ -0,0 +1,40 @@
...
+
+class PushRule extends BaseService {
+
+  create(projectId, options) {
+    const pId = encodeURIComponent(projectId);
+
+    return RequestHelper.post(this, `projects/${pId}/push_rule`, options);
+  }
+
+  async createOrEdit(projectId, options) {
```
casz: I can remove this if you consider this too much but I thought it be useful otherwise so people did not have to implement this all over.

jdalrymple:
Would it make sense to include the createOrEdit functionality within the edit function with an option?

jdalrymple:
OR have it included in the create function with an overwrite option?

casz:
Sounds reasonable 😅

jdalrymple:
lol which one makes MORE sense :P ?

casz:
Well create seems more reasonable since it throws error if it exists that more understandable then edit...

```
{
    "error": "Project push rule exists"
}
```

edit throws:

```
{
    "message": "404 Push Rule Not Found"
}
```

jdalrymple:
Sounds good to me! So just add an option to overwrite Wooo. This is awesome. Hopefully i can get it to release ASAP. Just waiting on the travis team for an update.

casz:
Just to be clear, your suggestion was removing createOrEdit and use overwrite option?

src/services/PushRule.js
```diff
@@ -0,0 +1,39 @@
+import { BaseService, RequestHelper } from '../infrastructure';
+
+class PushRule extends BaseService {
+
+  async create(projectId, options) {
```
jdalrymple: async create(projectId, options = {}) {

src/services/PushRule.js
```diff
@@ -0,0 +1,39 @@
...
+
+    if (options && options.overwrite) {
+      const push_rule = await this.show(projectId)
+      delete options.overwrite
+      if (push_rule) {
+        return this.edit(projectId, options)
+      }
+    }
+
+    return RequestHelper.post(this, `projects/${pId}/push_rule`, options);
```
jdalrymple: Would it be cleaner to do: 

```
...
const { overwrite, ...args } = options;

    if (overwrite) {
      const pushRule = await this.show(projectId);

      if (push_rule) return this.edit(projectId, args)
    }
    return RequestHelper.post(this, `projects/${pId}/push_rule`, args);
```

casz:
Overwrite would be delete and post
While
Update would be post or put

src/services/PushRule.js
```diff
@@ -34,6 +34,14 @@ class PushRule extends BaseService {
     return RequestHelper.get(this, `projects/${pId}/push_rule`);
   }
 
+  async update(projectId, options) {
```
casz: this avoid the options parameter and seems cleaner at the same time.

jdalrymple: I agree, my only reservation is the name, i feel like its close in meaning to edit no?

casz:
I like `upsert` as a wording for create or edit 👍 Perhaps upsert as a method instead of a parameter to avoid having edit as async function?

jdalrymple:
All of those functions are async, they just dont require the function prefix because the its contained in the return and doesnt include an `await` line :)

casz:
This reads horribly in edit 😞 I definitely prefer upsert just as an method
```javascript
  async edit(projectId, options) {

    const { upsert, ...args } = options;

    if (upsert) {
      const pushRule = await this.show(projectId);

      if (!pushRule) return this.create(projectId, args)
    }

    return RequestHelper.put(this, `projects/${pId}/push_rule`, args);
  }
```

jdalrymple:
This is a little cleaner

```javascript
  async edit(projectId, { upsert, ...options } ={}) {

    if (upsert) {
      const pushRule = await this.show(projectId);

      if (!pushRule) return this.create(projectId, options)
    }

    return RequestHelper.put(this, `projects/${pId}/push_rule`, args);
  }
```

Issue #137, opened 2018-07-31
TypeScript support


yonkeltron:
Hello,

I would love to use this library in a TypeScript project but it seems that [the only type definitions](https://www.npmjs.com/package/@types/gitlab) are third-party and woefully outdated. Any chance of getting official TypeScript support in the project?

Thanks,
+Jonathan

jdalrymple:
See #89, I'd like to finish moving things to ts, just running short on time lately :(.

yonkeltron:
I understand how it is to run short on time and I want to take this chance to thank you for your efforts thus far. Is there some pre-release branch which we can help test?

jdalrymple:
Its a huge team effort. Its wicked to see how everyone is pushing to make this library better. Yup! That is the current 'next' branch

jdalrymple:
I just rebased, there may be a few hiccups

damakawa:
Hello,

I'm trying to use the **next** branch, and I think I've found a few of the hiccups you mentioned.

Npm seems to be importing the module with no problems using the format `"gitlab": "git+https://github.com/jdalrymple/node-gitlab.git#next"` in the package.json dependencies.

However, when trying to import the module into a project file, the IDE isn't finding it in autocomplete, and adding it manually as `import * as GitLabApi from "gitlab";` is throwing the IDE error: 
```
[ts]
Could not find a declaration file for module 'gitlab'. '/Users/c74509/dev/TEA/gitrdonebot/node_modules/gitlab/dist/latest/index.js' implicitly has an 'any' type.
Try `npm install @types/gitlab` if it exists or add a new declaration (.d.ts) file containing `declare module 'gitlab';`
```

I went into the gitlabapi folder in node_modules and ran `npm run build`. The TS compiler is outputting the following errors:
```
src/infrastructure/XMLHttpRequester.ts:13:13 - error TS2348: Value of type 'typeof RequestHelper' is not callable. Did you mean to include 'new'?

13       await wait(sleepTime * 1000);
               ~~~~~~~~~~~~~~~~~~~~~~


src/services/MergeRequests.ts:101:26 - error TS2525: Initializer provides no value for this binding element and the binding element has no default value.

101   pipelines(projectId, { mergerequestId } = {}) {
                             ~~~~~~~~~~~~~~


src/services/Projects.ts:107:44 - error TS2304: Cannot find name 'Path'.

107   upload(projectId, filePath, { fileName = Path.basename(filePath) } = {}) {
[...]

PR #149, opened 2018-08-14
Fix wrong import

r00ger:
Thanks to #137, I tried to check ts build and compiler reported another runtime error


r00ger:
Btw, what's wrong with CI? Couldn't find the caller of failing `npm ci` method

jdalrymple:
Yea the npm ci call is coming from within Travis I think :( not sure what's going on there. I've been experimenting with downgrading libraries. I've also through of trying circle ci or even gitlab ci to see if it persists

PR #150, opened 2018-08-14
Make CI green again

r00ger:


src/services/PushRule.js
```diff
@@ -14,10 +13,10 @@ class PushRule extends BaseService {
     if (upsert) {
       const pushRule = await this.show(projectId);
 
-      if (!pushRule) return this.create(projectId, options)
+      if (!pushRule) return this.create(projectId, options);
     }
 
-    return RequestHelper.put(this, `projects/${pId}/push_rule`, args);
+    return RequestHelper.put(this, `projects/${pId}/push_rule`);
```
casz: This should be options, this is a put.

test/tests/services/PushRule.js
```diff
@@ -8,8 +8,8 @@ describe('PushRule', () => {
     });
     const pushRules = {
       upsert: true,
-      member_check: true
-    }
+      member_check: true,
+    };
```
casz: This is just a JSON literal does not need a semicolon. All you semicolon lovers 😅

src/services/PushRule.js
casz: This should be options as second argument since it is a put.

test/tests/services/PushRule.js
r00ger: Well, normally your IDE should pick up `eslint` config and blame you for not following the standard.

Anyway, who cares, let's just make CI happy and green

casz: well my vscode did not pickup any eslint errors, did not have the extension installed 😢

r00ger:
@casz can you please take a look at the [failing PushRule test](https://travis-ci.org/jdalrymple/node-gitlab/jobs/416004245)?

If you don't have time at the moment, I'd suggest to disable the test and proceed w/ merge, so that CI can start revealing problems in pending pulls

r00ger:
Oh, wait a second. Did it ever work? It should not work w/o licence uploaded, since it a EE feature

https://gitlab.com/gitlab-org/gitlab-ee/issues/3825

casz:
Ehh sorry did not think of that @r00ger my bad 😭

test/tests/services/PushRule.js
casz: Maybe setup pre-commit hook to run "lint" script?

r00ger:
@jdalrymple 🎉

casz:
caching can be evil 😈

PR #153, opened 2018-08-15
fix next branch linting and renaming files

casz:
Without `--project tsconfig.json` in npm lint there was a warning, see https://github.com/palantir/tslint/issues/3864


.travis.yml
```diff
@@ -3,7 +3,8 @@ language: node_js
 stages:
   - lint and build
   - test
-  - release
+  - name: release
+    if: branch = master
```
casz: Don't know if this is wanted but I could not explain why it was unable to find `@semantic-release/git` module

PR #157, opened 2018-08-15
fix Master

casz:
fix linting after https://github.com/jdalrymple/node-gitlab/commit/bc3b366ac69abdab10ed3c002dd07632a7b9a044

PR #159, opened 2018-08-15
Fix linter failures in RequestHelper

r00ger:


casz:
@r00ger was already fixed in #157 so a rebase should be good enough

PR #164, opened 2018-08-16
Support rejectUnauthorized parameter

r00ger:


jdalrymple:
lol I keep messing up the rebasing :upside_down_face:

r00ger:
Thanks. In that case, you haven't missed anything. Class' prop has to be specified explicitly in TS, so few lines have to be added anyway

PR #185, opened 2018-09-06
Add types

r00ger:
In this PR I want to add basic types so that we can emit declarations

### DIFF
--- a/src/infrastructure/BaseService.ts
+++ b/src/infrastructure/BaseService.ts
@@ -17,10 +17,10 @@ export type BaseModelContructorOptions =
   | BaseModelOptions & Required<Pick<BaseModelOptions, 'oauthToken'>>;
 class BaseModel {
   protected url: string;
-  public headers: { [header: string]: string | number};
-  public rejectUnauthorized: boolean;
-  protected requester: any;
-  protected useXMLHttpRequest: boolean;
+  public readonly headers: { [header: string]: string | number};
+  public readonly rejectUnauthorized: boolean;
+  protected readonly requester: any;
+  protected readonly useXMLHttpRequest: boolean;
 
   constructor({
     token,
--- a/src/infrastructure/BaseService.ts
+++ b/src/infrastructure/BaseService.ts
@@ -12,6 +12,9 @@ interface BaseModelOptions {
   rejectUnauthorized?: boolean;
 }
 
+export type BaseModelContructorOptions =
+  | BaseModelOptions & Required<Pick<BaseModelOptions, 'token'>>
+  | BaseModelOptions & Required<Pick<BaseModelOptions, 'oauthToken'>>;
 class BaseModel {
   protected url: string;
   public headers: { [header: string]: string | number};
@@ -19,8 +22,6 @@ class BaseModel {
   protected requester: any;
   protected useXMLHttpRequest: boolean;
 
-  constructor(options: BaseModelOptions & Required<Pick<BaseModelOptions, 'token'>>);
-  constructor(options: BaseModelOptions & Required<Pick<BaseModelOptions, 'oauthToken'>>);
   constructor({
     token,
     oauthToken,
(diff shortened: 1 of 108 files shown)


src/infrastructure/BaseService.ts
```diff
@@ -11,23 +11,24 @@ interface BaseModelOptions {
   rejectUnauthorized?: boolean;
 }
 
+export type BaseModelContructorOptions =
```
jdalrymple: Should these options be in their own directory ?

r00ger: It's a matter of taste. I personally, keep types in the same file as long as it doesn't affect readability

jdalrymple: I would keep them in if they are relatively small, but i feel like the types information will start growing pretty quickly. It may be easier to read if they were all located together instead of being embedded in the file.

casz:
These seem pretty feature complete at least for the types for node-gitlab module.
Where you planning to also add return types for functions? So developers could use those types in their own projects?

r00ger:
@jdalrymple how come `codecov` failed the build? What's the criteria to pass?

r00ger:
@jdalrymple @casz please review the changes

src/templates/ResourceMembers.ts
```diff
@@ -1,47 +1,51 @@
...
     this.url = URLJoin(this.url, resourceType);
   }
 
-  all(resourceId, includeInherited = false, options = {}) {
+  all(resourceId: ResourceId, includeInherited = false, options = {}) {
     const rId = encodeURIComponent(resourceId);
 
     const url = includeInherited ? `${rId}/members/all` : `${rId}/members`;
 
-    return RequestHelper.get(this, url, { options });
```
r00ger: `{ options }` here seems to be an error

src/templates/ResourceAccessRequests.ts
```diff
@@ -12,34 +13,34 @@ export const ACCESS_LEVELS = {
...
     return RequestHelper.post(this, `${rId}/access_requests`);
   }
 
-  approve(resourceId, userId, { accessLevel = 30 }) {
+  approve(resourceId: ResourceId, userId: UserId, { accessLevel = 30 }) {
     const [rId, uId] = [resourceId, userId].map(encodeURIComponent);
 
     return RequestHelper.post(this, `${rId}/access_requests/${uId}/approve`, {
-      accessLevel,
+      access_level: accessLevel,
```
casz: Seems like a mistake when we have https://github.com/jdalrymple/node-gitlab/blob/f946569882bdba81bda1e8e34c1c7c5d968fbd70/src/infrastructure/RequestHelper.ts#L46-L55

src/infrastructure/RequestHelper.ts
```diff
@@ -116,8 +124,74 @@ async function getPaginated(service, endpoint, options: GetPaginatedOptions = {}
...
+  expiresAt?: string;
+  note?: string;
+  actions?: CommitAction[];
+  commitMessage?: string;
+  branch?: string;
+  body?: string | temporaryAny;
+  title?: string;
+  name?: string;
+  label_id?: temporaryAny;
+  access_level?: number;
```
casz: this should be `accessLevel`

src/infrastructure/RequestHelper.ts
```diff
@@ -116,8 +124,74 @@ async function getPaginated(service, endpoint, options: GetPaginatedOptions = {}
...
+  user_id?: UserId;
```
casz: this should be userId

casz:
@r00ger do think it would be possible to add Promise<GitLabProject> or similar where it makes sense on methods that have return type?

casz:
@r00ger sounds like a plan 😅

Could you resolve merge conflicts? 🤭

casz:
@jdalrymple I think this is mergable 😅

PR #189, opened 2018-09-16
Added user snippets API support

brunobastosg:
User snippets do not belong to a specific project and have a [different API resource](https://docs.gitlab.com/ee/api/snippets.html)  than the project snippets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ NotificationSettings
 PagesDomains
 Search
 SidekiqMetrics
+Snippets
 SystemHooks
 Version
 Wikis
--- a/src/services/Snippets.ts
+++ b/src/services/Snippets.ts
@@ -1,0 +1,53 @@
+import { BaseService, RequestHelper } from '../infrastructure';
+
+class Snippets extends BaseService {
+  all(options = {}) {
+    return RequestHelper.get(this, `snippets`, options);
+  }
+
+  allPublic(options = {}) {
+    return RequestHelper.get(this, `snippets/public`, options);
+  }
+
+  content(snippetId) {
+    const sId = encodeURIComponent(snippetId);
+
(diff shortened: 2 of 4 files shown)


jdalrymple:
Don't know how i missed that lol. In your PR you have a content function pointing to a raw endpoint, but i dont see that endpoint in the docs? Also you could probably combine all and allPublic and add public as an optional param like:

```javascript
...
  all(options = { public: false }) {
    const { public, ...opts } = options
    const url = public ? 'snippets/public' : 'snippets';

    return RequestHelper.get(this, url, opts);
  }
```

To cut down on the amount of functions.

brunobastosg:
The `raw` endpoint is not in the docs, but it exists. For example, https://gitlab.com/snippets/1754319/raw.

I combined `all` and `allPublic` as you suggested, and also added the Snippets API to the README.

jdalrymple:
Looks good!!

Just a couple things left! 

- Could you adjust that one lint error, i think its switching ` to ' on line 16
- If this is a user specific snippet should it be added to the user bundle ? As in when someone wants to import all the user API's such as emails, or keys etc, should this be included? If yes, we will need to add it to the bundle and update this test https://github.com/jdalrymple/node-gitlab/blob/master/test/tests/bundles/UsersBundle.ts

brunobastosg:
Fixed the lint error.

I don't know if it should be added to the user bundle. There is no `/users` anywhere in the url. The snippets are identified only by their id.

It's up to you. If you really want it I can add it.

jdalrymple:
That makes sense. We'll leave it out.

brunobastosg:
Alright then.

Should I add some tests to fix that codecov check?

jdalrymple:
If you have some time that would be amazing!

PR #220, opened 2018-10-17
services/NotificationSettings: fix #219

### LINKED ISSUE #219: Global NotificationSettings not working
## Description
calling NotificationSettings.edit or .all does not work on global settings because of [NotificationSettings.ts#L54](https://github.com/jdalrymple/node-gitlab/blob/master/src/services/NotificationSettings.ts#L54)

## Steps to reproduce
```javascript
const Gitlab = require('gitlab/dist/es5').default
const gitlab = new Gitlab({
  url: 'your url here',
  token: 'your token here'
})
async function run () {
  await gitlab.NotificationSettings.edit({
    level: gitlab.NotificationSettings.LEVELS.DISABLED
  })
}
run()
```
## Expected behaviour
global NotificationSettings should be updated

## Actual behaviour
Error because the url is not set properly
```
name: 'StatusCodeError',
  statusCode: 404,
  message: '404 - {"error":"404 Not Found"}',
  error: { error: '404 Not Found' },
  options:
   { url:
      'http://gitlab.yoururlhere.com/api/v4/undefinednotification_settings',
     headers: { 'private-token': 'your private token here' },
     json: true,
     body: { level: 'disabled' },
     resolveWithFullResponse: false,
     rejectUnauthorized: true,
     method: 'PUT',
     callback: [Function: RP$callback],
     transform: undefined,
     simple: true,
     transform2xxOnly: false },
```
the url is `api/v4/undefinednotification_settings` which obviously shows that the var `url` is not defined when no project or group has been declared.

## Possible fixes
change `let url` to `let url = ''` on [NotificationSettings.ts#L54](https://github.com/jdalrymple/node-gitlab/blob/master/src/services/NotificationSettings.ts#L54) and [NotificationSettings.ts#L42](https://github.com/jdalrymple/node-gitlab/blob/master/src/services/NotificationSettings.ts#L42)

This call is perfectly valid as per the [API Documentation#Update global notification settings](https://docs.gitlab.com/ce/api/notification_settings.html#update-global-notification-settings)

Musinux:
see #219

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@
         - [Bundle Imports](#bundle-imports)
     - [Examples](#examples)
     - [Pagination](#pagination)
+    - [sudo](#sudo)
 - [Migrating from node-gitlab](#migrating-from-node-gitlabnode-gitlab)
 - [Docs](#docs)
 - [Development](#development)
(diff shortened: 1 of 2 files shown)


jdalrymple:
Nice catch!

PR #221, opened 2018-10-17
doc: explain Sudo header

Musinux:
Hello,
I had a hard time understanding how to be able to impersonate users via the API from your module, so I thought I'll write some doc about it for others to be able to do it the same way.

I hope it will be useful !

### DIFF
--- a/README.md
+++ b/README.md
@@ -386,19 +386,15 @@ For example, if you want to disable notifications for a specific user:
 ```javascript
 import Gitlab from 'gitlab';
 
-const api = new Gitlab({
+const { NotificationSettings } = new Gitlab({
   url:   'http://example.com', // Defaults to http://gitlab.com
   token: 'abcdefghij123456' // Can be created in your profile.
+  sudo: userid_or_username
 });
-
-api.NotificationSettings.headers.Sudo = userid_or_username; // eg: 1 or 'username'
 
 await api.NotificationSettings.edit({
   level: api.NotificationSettings.LEVELS.DISABLED
 })
-
-delete api.NotificationSettings.headers.Sudo // clear impersonation header afterwards
-
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -378,6 +379,28 @@ pagination: {
...
 ```
 
+### Sudo
+For private gitlab instances, administrators are able to impersonate users through the API. To do so, you have to set the 'Sudo' header on the services you want to impersonate the user for.
+
+For example, if you want to disable notifications for a specific user:
+```javascript
+import Gitlab from 'gitlab';
+
+const api = new Gitlab({
```
jdalrymple: Id suggest changing this to:

```javascript

const { NotificationSettings } = new Gitlab({
  url:   'http://example.com', // Defaults to http://gitlab.com
  token: 'abcdefghij123456' // Can be created in your profile
  sudo: userid_or_username
}) ;

....
```

Or if youre looking to add sudo for a specific api request, i should expose it differently then having to add/delete the prop from the headers.

jdalrymple: If so, ill make a MR tonight to fix this for the next version so it will be much clearer. I really want to avoid changing the internals of the services once they've been initialized.

Musinux: Nevermind, I just saw the merged [PR](https://github.com/jdalrymple/node-gitlab/pull/210/commits/4bf574c0c019fe42f6d77aab15d558b4cfbb321a) about the sudo header. I only checked my local version of the code (4.1.1) and not the master branch.

jdalrymple: The current release assumes that you init it with the service, but if youd rather have it on a singular call vs on every call of that service, i can make a PR to handle that in a cleaner fashion :)

Musinux: Well, I'm not against a cleaner way to do this !

jdalrymple: Created an example here #223

Issue #98, opened 2018-06-04
Using XMLHttpRequests results in broken 'fs' dependency


icosamuel:
I'm trying to use your api client [in the browser](https://github.com/jdalrymple/node-gitlab/issues/56). So far i've been unable to build my SPA because of a dependency thats not available in the browser (fs). 

```
$ npm install --save fs
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.1.3 (node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.1.3: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})

+ fs@0.0.1-security
updated 1 package in 17.193s
```

```
This dependency was not found:

* fs in ./node_modules/gitlab/dist/latest/services/Projects.js, ./node_modules/request/lib/har.js

To install it, you can run: npm install --save fs
```
It seems ratter normal that fs cannot be used in the browser [[1]](https://github.com/nodejs/help/issues/547) [[2]](https://stackoverflow.com/questions/50370975/npm-stream-player-causing-fs-dependency-not-found).

What can I do to work around this?

If no workarounds are possible, Is there anyone else encountering the same issue and should we try to make a browser-friendly version?

jdalrymple:
Wait the standard node fs library isn't available but the other ones are ? :P something about that is off

jdalrymple:
Hmm. Lemme look

jdalrymple:
It's only used for the upload function, to read in files. Could switch it with another library but I'm not sure that would solve your problem

icosamuel:
There is also a reference in the `request` module that acts as a dependency to this project. I'm not sure either how to go about solving this

jdalrymple:
Me neither :s I'll do some research tonight

icosamuel:
So I've spent a couple of hours migrating the http-request lib to Axios. In the process, I eliminated all dependencies to `fs`. So I guess this would count as option 1, but not easily mergable.

I've forked the project to meet my needs, but I'd love for some ideas on how to contribute to your project. Here it is: https://github.com/icosamuel/node-gitlab/pull/1

jdalrymple:
@Mr-Wallet any suggestions?

Mr-Wallet:
[...]

PR #223, opened 2018-10-18
Adding the ability to add `sudo` to specific requests

jdalrymple:
## Summary

```javascript
const { NotificationSettings } = new Gitlab({
  url:   'http://example.com', // Defaults to http://gitlab.com
  token: 'abcdefghij123456' // Can be created in your profile
}) ;

await NotificationSettings.edit({
  level: NotificationSettings.LEVELS.DISABLED
  sudo: 8,
})
```


jdalrymple:
This will be added once #98 is finished

casz:
Cleaner diff with the current base branch https://github.com/jdalrymple/node-gitlab/compare/remove-dependency-on-fs...specific-sudo

jdalrymple:
Probably not for another week or so? Depends on how much progress i make on the other issue :)

Musinux:
Okay, thanks !

PR #225, opened 2018-10-24
fix the bug of Paginated

hucheng91:
when i  use GroupProjects.all(id),  and the projects have three pagegs, when Paginated url have a problem

 example:
 ```javascript
  const values =  gitlabApi.GroupProjects.all(id); // there have 60 projects
 
 when paginated 3 page 
url = ""http://xxx.net/api/v4""
endpoint = "https://xxx.net/api/v4/groups/123/projects?archived=false&id=123&order_by=created_at&owned=false&page=2&per_page=20&simple=false&sort=desc&starred=false"
 ```

so I edit it

### DIFF
--- a/src/infrastructure/RequestHelper.ts
+++ b/src/infrastructure/RequestHelper.ts
@@ -33,8 +33,14 @@ function defaultRequest(
   endpoint,
   { headers, body, qs, formData, resolveWithFullResponse = false }: RequestParametersInput,
 ): RequestParametersOutput {
+  let paramUrl;
+  if(endpoint.startsWith("http")){
+    paramUrl = endpoint;
+  }else{
+    paramUrl =  URLJoin(url, endpoint)
+  }  
   const params: RequestParametersOutput = {
-    url: URLJoin(url, endpoint),
+    url: paramUrl,
     headers,
     json: true,
   };


jdalrymple:
So your url is : `http://xxx.net/api/v4`, but Gitlab is returning `https://xxx.net/api/v4/groups/123/projects?archived=false&id=123&order_by=created_at&owned=false&page=2&per_page=20&simple=false&sort=desc&starred=false` for the next page url?

casz:
I would not accept this change.
If https is possible, USE IT! Matter of importing CA if that's the case.
https://nodejs.org/api/cli.html#cli_node_extra_ca_certs_file

the extra ca certs file can be a pem with multiple certificates included.

hucheng91:
Oh ,Sorry, My English is so pool

the problem is  if one group have more than 20 projects (eg: 80)

```javascript
when page=1&per_page=20

url = "https://git.qutoutiao.net/api/v4"
endpoint = "groups/123/projects"

URLJoin(url, endpoint) => "https://git.qutoutiao.net/api/v4/groups/123/projects"  (is good)

but when page=2&per_page=20

url = "https://git.qutoutiao.net/api/v4"
endpoint = "https://git.qutoutiao.net/api/v4/groups/123/projects?archived=false&id=123&order_by=created_at&owned=false&page=2&per_page=20&simple=false&sort=desc&starred=false"

URLJoin(url, endpoint) =>  "https://git.qutoutiao.net/api/v4/https://git.qutoutiao.net/api/v4/groups/123/projects?archived=false&id=123&order_by=created_at&owned=false&page=2&per_page=20&simple=false&sort=desc&starred=false"  (is bad)
so I change the code  
I want when the page = 2,3,4,5,6,......,not use URLJoin

```

jdalrymple:
Ah! I understand. I had this problem before as well, not sure how it returned. Its due to one line in the getPaginate function. Ill look at it in the next day and push the fix. Thank you for finding this!

jdalrymple:
Will be fixed in the next release (Version 5!) Sorry for the delay :(

PR #228, opened 2018-10-30
Fix unreliable application settings test

r00ger:
it tests internals of Gitlab instead of funtionality, and fails on new versions of Gitlab, if they change the list of available settings


jdalrymple:
Much better idea, thanks!

Issue #234, opened 2018-11-15
XMLHttpRequester is deleting resources on a GET


Mr-Wallet:
XMLHttpRequester's `get` method is running a `del` on the internal XHR implementation. This was introduced in commit [ffad9a3](https://github.com/jdalrymple/node-gitlab/commit/ffad9a34dae5e9ffb19f97666e5653ded7e28551#diff-576e56cbae02978fa2d8246b029c2075R45)

PR #235, opened 2018-11-15
fixes: #234 use correct XMLHttpRequest for GET req

casz:
There @Mr-Wallet 😎


Mr-Wallet:
Thanks @casz, that was fast!

PR #244, opened 2018-11-28
WIP: Add unit tests to next

beaverusiv:
This is an experiment to add unit tests to the repository. I see you have setup rigging for integration testing which is nice but I am of the opinion that unit testing should not be left out.

The big change needed is to make `RequestHelper` non-static so that we can mock it. This current batch of tests look pretty trivial and a great many will be but they still come together to provide a safety net for changes and even help development when the Gitlab API changes as you can change the unit tests first to expect the new formats.

As you can see at `src/services/Projects.ts:31` adding the tests already exposed the fact that `userId` was a required parameter when it should be optional.

I'm not expecting to have this accepted as-is but I am looking for feedback as to if this is something that is welcome so I can move forward.


casz:
This is super awesome 😍

jdalrymple:
Looks good! 

Is there a reason why you decided to go for mocking the RequestHelper instead of the got library?

beaverusiv:
Unit tests should only test a single unit (function) so this is best practice. When I test `RequestHelper` itself then it will be `got` being mocked out.

jdalrymple:
Makes sense! Last question, is there a reason why it must be made into non static functions to be mocked?

beaverusiv:
afaik yes, because we need to be able to instantiate a mocked version for use in the tests. If you know of another way then by all means, we can try that. I would prefer to do as little changing of the current code as possible.

You can pull down this branch and run the tests in your IDE. The reason the CI fails is because compilation fails as the other services haven't been modified to work with this version of `RequestHelper`.

jdalrymple:
Ill give it a look in the next couple days and see what i can do! Looks good though :) Thank you!!

jdalrymple:
I think i got it working. Though i would like to seperate out the unit tests from the integration tests, ill make a branch as a proof of concept

jdalrymple:
Yea i thought about that initially, i just didnt like the way the structure looked lol. Having them in their own folder looked cleaner to me :joy:

beaverusiv:
It's a lot more convenient when developing to have them side-by-side but it is a stylistic choice essentially. If you want to create a new branch and play with I can take the reigns once I've done the documentation if I need to to complete it. I feel very strongly about both documentation and code coverage.

jdalrymple:
:sunglasses: Love hearing that! Ill finish it up tonight and pass it off!

jdalrymple:
I just merged my changes!

beaverusiv:
Looks good, I'll pull down and play around with it today when I do some more work on the documentation.

jdalrymple:
How is it any more difficult to mock as a plain function then a class? I really want to avoid having it as a property of each service if it can be avoided :(

beaverusiv:
I couldn't get it to work nicely, the mocking was getting in the way of the Projects import.

https://en.wikipedia.org/wiki/Inversion_of_control

jdalrymple:
Not sure i understand what you mean about the Projects import?

beaverusiv:
When I tried mocking out `RequestHelper` as an object with functions it would mock but then `Projects` would import as an empty object `console.debug(service) // => '{}'`

jdalrymple:
Ah! Weird, Ill look into it tomorrow, otherwise well keep it as a property of the service like youve done

jdalrymple:
Fixed, give it a look when you get a chance. BigInt is supported as well in the most recent version of node/typescript: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-2.html

beaverusiv:
I'll leave this up to you as I've got my hands full with the docs. Even if it only gets merged with the few tests I put in it is still more than before, and should make it easier for someone to extend and add more coverage.

jdalrymple:
I added whats already been done. Going to keep the issue and MR open until more tests are written, but ill focus on getting 5.0.0 out for now.

PR #248, opened 2018-12-12
feat(services): add protected tags

Neonox31:
Add protected tags endpoint described here : https://docs.gitlab.com/ee/api/protected_tags.html#list-protected-tags


jdalrymple:
Looks good!

PR #216, opened 2018-10-12
Version 5.0.0

jdalrymple:
# Summary
## Breaking
* Added content as a required parameter for RepositoryFiles
* Removed projectId from System Hooks API since it wasn't required
* Removed dependency on FS. Now the Projects API takes in two arguments `projectId` and `content` as well as an option fileName argument
* Changing everything to named exports for simplicity
* Switching required initialization argument from 'url' to 'host'
* Updated Approvals API support to match https://docs.gitlab.com/ee/api/merge_request_approvals.html
* MergeRequest Pipelines require the mergeRequestId
* NotificationSettings API edit function now takes one parameter, `options`
* Changing the access level enum property from master to maintainer as per https://gitlab.com/gitlab-org/gitlab-ce/issues/42751
* Notes now require a body argument instead of checking the options argument for a body parameter

## Bug Fixes
* #227 (https://github.com/jdalrymple/node-gitlab/issues/227) Fixing array syntax thanks to Lukas Eipert (https://github.com/leipert) (f9bc34d (https://github.com/jdalrymple/node-gitlab/commit/f9bc34d))
* Fixing Todos support. If todoId was not passed, an undefined value would be introduced into the url (cbeef18 (https://github.com/jdalrymple/node-gitlab/commit/cbeef18))
* ResourceAwardEmojis API wasn't properly filtering based on awardId (70f4315 (https://github.com/jdalrymple/node-gitlab/commit/70f4315))
* Removed xhr library in favour of ky, and switched request for got for a smaller package size and retry functionality

## Features
* Added the ability to add sudo to specific requests (780244f (https://github.com/jdalrymple/node-gitlab/commit/780244f))
* Added the missing edit function to the Groups API
* Added LDAP support to the Groups API 

## Documentation
* Removing xml request docs
* Updating imports to be named imports

PR #250, opened 2018-12-24
WIP: Try implementation only with `ky`

LoicMahieu:
Related to #216
Just a POC. Works in my use-case.

### DIFF
--- a/src/infrastructure/BaseService.ts
+++ b/src/infrastructure/BaseService.ts
@@ -1,4 +1,4 @@
-import Request from 'got';
+import Request from 'ky';
 import { BaseServiceOptions } from '../../types/types';
 
 export class BaseService {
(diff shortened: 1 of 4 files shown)


jdalrymple:
With the switch, does this also support node or just the browser implementation?

sindresorhus:
FYI, someone is working on getting Ky to work in non-DOM environments: https://github.com/sindresorhus/ky/pull/69 So you could use it with [`node-fetch`](https://github.com/bitinn/node-fetch) to get support for Node.js. Try it out and help review to get it landed faster ;)

This is how you get Ky to work in Node.js currently: https://github.com/sindresorhus/ky/blob/master/test/_require.js When the mentioned PR is merged, line 9-11 can be removed.

LoicMahieu:
Close to #265 that use Ky from recents changes from next

PR #259, opened 2019-01-29
Update README.md with `rejectUnauthorized`

EvHaus:
After upgrading my version of `node-gitlab` I noticed that I the API broke as result of these error messages: `Error: self signed certificate`. I was really confused why this was happening even though I had `process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';` set in my file.

Turns out this work https://github.com/jdalrymple/node-gitlab/issues/142 added a hidden flag to Gitlab's BaseModel to specifically handle rejections of certificates. I couldn't find any existing documentation for it.

This PR adds documentation for it so that others running into this issue won't waste their time going down the same rabbit hole I went on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,8 +148,9 @@ const Gitlab = require('gitlab/dist/es5').default
 
 // Instantiating
 const api = new Gitlab({
-  url:   'http://example.com', // Defaults to https://gitlab.com
-  token: 'abcdefghij123456'	// Can be created in your profile.
+  url:   'http://example.com',  // Defaults to https://gitlab.com
+  token: 'abcdefghij123456',    // Can be created in your profile.
+  rejectUnauthorized: false     // Should insecure certificates be rejected (true by default)
 })
 
 // Or, use a OAuth token instead!
--- a/README.md
+++ b/README.md
@@ -148,9 +148,8 @@ const Gitlab = require('gitlab/dist/es5').default
 
 // Instantiating
 const api = new Gitlab({
-  url:   'http://example.com',  // Defaults to https://gitlab.com
-  token: 'abcdefghij123456',    // Can be created in your profile.
-  rejectUnauthorized: false     // Should insecure certificates be rejected (true by default)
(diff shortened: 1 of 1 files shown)


casz:
again, this rabbit hole is easily solved if you would import your certificates.
I cannot say enough how much you leave yourself open to all the nastiness by allowing unauthorized TLS connections in your infrastructure.

https://github.com/jdalrymple/node-gitlab/issues/142#issuecomment-412304991

README.md
```diff
@@ -148,8 +148,9 @@ const Gitlab = require('gitlab/dist/es5').default
 
 // Instantiating
 const api = new Gitlab({
-  url:   'http://example.com', // Defaults to https://gitlab.com
-  token: 'abcdefghij123456'	// Can be created in your profile.
+  url:   'http://example.com',  // Defaults to https://gitlab.com
+  token: 'abcdefghij123456',    // Can be created in your profile.
+  rejectUnauthorized: false     // Should insecure certificates be rejected (true by default)
```
casz: Please document it separately away from Instantiating example and include mention of [`NODE_EXTRA_CA_CERTS`](nodejs.org/api/cli.html#cli_node_extra_ca_certs_file)

EvHaus: Created a whole new section for describing HTTPS handling. How's that?

casz: Thanks!

README.md
```diff
@@ -268,6 +268,28 @@ EpicNotes
...
+},
+```
+
+Although we don't encourage it, if you absolutely must allow insecure certificates, you can instantiate the API with `rejectAuthorized` set to `true` like this:
+
+```
+const api = new Gitlab({
+  url: '...',
+  token: '...',
+  rejectUnauthorized: true
```
casz: whoops, should this not be set to `false` 😅

EvHaus: Doh! Thanks for that. Will push an update shortly.

PR #272, opened 2019-02-20
fix: using form-data

Vrtak-CZ:
see: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/form-data/index.d.ts#L18 and https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/form-data/index.d.ts#L32


casz:
@Vrtak-CZ this is most likely fixed in the [next branch](../tree/next)

Vrtak-CZ:
please merge https://github.com/jdalrymple/node-gitlab/pull/271 first

PR #298, opened 2019-04-09
Dropped node version back to 8.9.0 while it is still in Maintenance LTS

bigkraig:
This came up in https://github.com/danger/danger-js/pull/855 where we are adding GitLab support to Danger JS. I didn't see any specific reason to drop support for 8.x which is still supported until Dec 2019.


casz:
Please update Travis to build against 8.9.0 as well.

bigkraig:
@casz any idea whats going on with travis? I am not very familiar with how to configure this but I assumed adding another version of node wouldn't create new jobs that also ran the integration tests like that.

bigkraig:
@casz all set!!

jdalrymple:
I'm going to check and see if we can move the two version tests into their respective stages. Not sure if it will work tbh, but if it doesnt, ill just revert it.

casz:
@jdalrymple it is possible :)

Using matrix includes https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/.travis.yml

jdalrymple:
@casz That method gets rid of the stages though :(. I found this: https://github.com/travis-ci/travis-ci/issues/8295 ?

jdalrymple:
So i added the modifications, but the integration tests seem to fail for 8.9.0. Not sure why exactly :s

bigkraig:
> So i added the modifications, but the integration tests seem to fail for 8.9.0. Not sure why exactly :s

I think we just need to run the docker-compose as part of the 8.9.0 integration test, will know shortly.

bigkraig:
@casz @jdalrymple all green! :+1:

jdalrymple:
So the before_script part also had to be duplicated ? That's annoying lol. Looks good though! Ill merge and release tonight

casz:
I think you can clean up the travis file even more :sweat_smile: Let me try it out.

casz:
Nope :cry:

jdalrymple:
Had my hopes up there for a moment @casz

casz:
So did I :disappointed:

PR #301, opened 2019-04-11
Add maxPages to RequestOptions

jtheoof:
This field is used during pagination requestions for `all` functions.


jdalrymple:
Thanks! This has been fixed in the next branch already aswell!

PR #304, opened 2019-04-23
add Markdown in supported API's features

gpineda-dev:
# context

This is a minimalist Pull but i was looking for [markdown rendering](https://docs.gitlab.com/ee/api/markdown.html) (by gitlab api) and i discovered you have implemented it ! So it's really interesting, thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Keys
 Licence
 LicenceTemplates
 Lint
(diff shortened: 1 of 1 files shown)


jdalrymple:
Thanks for updating the docs! I tend to miss things :confused:

Issue #313, opened 2019-05-15
Next: new, breaking camelizeKeys behavior is not documented and could be optional?


krische:
**Description**
I noticed the new `KyRequester` functionality is running [camelizeKeys](https://github.com/jdalrymple/node-gitlab/blob/a999545542883803b13f4e8e981c49633a69d738/src/infrastructure/KyRequester.ts#L50) on the entire response body. This seems, to me, a very significant change, but I didn't see it mentioned anywhere in the documentation or the change log.

Perhaps there should even be an option to enable/disable that behavior? I could see that change requiring a lot of refactoring for some users.

jdalrymple:
It probably shouldn't be there, at least not without an option. Good catch !

jdalrymple:
Found another problem while making the switch back. Ill get to this before the weekend :)

jamime:
Fixed by https://github.com/jdalrymple/node-gitlab/commit/ec59c791970ef0249365288343e8c3119f7fa53f

jdalrymple:
Yea I havent published the fix yet. Still awaiting on some guidance on this question: https://github.com/jdalrymple/node-gitlab/pull/216#issuecomment-492870154

jdalrymple:
Made the changes, just testing now. It will be optional :)

jdalrymple:
Published in next :)

PR #316, opened 2019-05-22
feat: Adding the ability to camelizekeys of response body #313

jdalrymple:
Adding the ability to camelize the response body optionally


jdalrymple:
Not sure why the tests are failing intermittently  :s

PR #350, opened 2019-06-11
344 pagination url

jdalrymple:


casz:
@jdalrymple this regex handles the case of service url providing `/groups` or `/projects` etc... it also handles the case if the leaf is `v4`
leaf is the last thing after the last slash :sweat: 
```javascript
const leaf = service.url.split('/').pop()
new RegExp(`.+\/api\/v\d(\/${leaf})?\/`)
```

https://regex101.com/r/2LSP9L/1

casz:
I'll see if I can send a PR, hopefully with a test

PR #353, opened 2019-06-12
fix: pagination url - use regex to remove service url + leaf

casz:
Hmm couldn't figure out a good way to add a test :sweat:

### DIFF
--- a/src/infrastructure/RequestHelper.ts
+++ b/src/infrastructure/RequestHelper.ts
@@ -40,7 +40,9 @@ export async function get(
   // Rescurse through pagination results
   if (!query.page && underLimit && pagination.next) {
     const { next } = Li.parse(headers.link);
-    const more = await get(service, next.replace(`${service.url}/`, ''), {
+    const leaf = service.url.split('/').pop() || ""
+    const regex = new RegExp(`.+\/api\/v\d(\/${leaf})?\/`)
+    const more = await get(service, next.replace(regex, ''), {
       maxPages,
       sudo,
       showPagination: true


src/infrastructure/RequestHelper.ts
```diff
@@ -40,7 +40,9 @@ export async function get(
   // Rescurse through pagination results
   if (!query.page && underLimit && pagination.next) {
     const { next } = Li.parse(headers.link);
-    const more = await get(service, next.replace(`${service.url}/`, ''), {
+    const leaf = service.url.split('/').pop() || ""
```
jdalrymple: if the url was http://gitlab.com/api/v1, wouldnt leaf be v1?

jdalrymple:
> Hmm couldn't figure out a good way to add a test sweat

Hmm maybe by mimicking the pagination tests with different urls ?

src/infrastructure/RequestHelper.ts
casz: The regex handles it just fine:

See here: https://regex101.com/r/2LSP9L/1

casz:
> Hmm maybe by mimicking the pagination tests with different urls ?

Feel free to add the tests, my mocking skills are not that good :sweat: Much better at regex :laughing:

src/infrastructure/RequestHelper.ts
jdalrymple: In that case leaf is projects, but what if leaf were v1, like the case I pointed about above.

casz: My dear sir, you can modify the regex :sweat:

jdalrymple: Lol I realised that after I commented.

jdalrymple:
Ill add tests later, this is gonna cause many people problems if its now pushed ASAP

casz:
hold off :sweat:

casz:
it is missing a single backlash for digits... because it is a string regex :cry: 

I just tested it in a playground.

Issue #254, opened 2019-01-04
Pipeline.create not work with .variables


MrCheater:
**Description**
version: "5.0.0-rc.7"

```
const PipelinesService = new Gitlab.Pipelines({
  token: 'XXX-XXX-XXX'
})

await PipelinesService.create(gitlabProjectId, 'ci/cd', {
  variables: {
    PULL_REQUEST_NAME: `${pullRequest.title}`,
    PULL_REQUEST_NUMBER: `${pullRequest.number}`,
    HEAD_COMMIT_SHA: `${pullRequest.head.sha}`,
    MERGE_COMMIT_SHA: `${pullRequest.merge_commit_sha}`,
    SOURCE_BRANCH_NAME: `${pullRequest.head.ref}`,
    SOURCE_BRANCH_SHA: `${pullRequest.head.sha}`,
    TARGET_BRANCH_NAME: `${pullRequest.base.ref}`,
    TARGET_BRANCH_SHA: `${pullRequest.base.sha}`
  }
})
```

Error:

```
{ HTTPError: Response code 500 (Internal Server Error)
    at EventEmitter.emitter.on (/home/mrcheater/resolve/node_modules/got/source/as-promise.js:74:19)
    at process.internalTickCallback (internal/process/next_tick.js:77:7)


{"ref":"ci/cd","variables":{"p_u_l_l__r_e_q_u_e_s_t__n_a_m_e":"Feature/Transactional read models","p_u_l_l__r_e_q_u_e_s_t__n_u_m_b_e_r":"950","h_e_a_d__c_o_m_m_i_t__s_h_a":"d5c03dca7b2688f914eec50382b064d63ac9caad","m_e_r_g_e__c_o_m_m_i_t__s_h_a":"cacfdc22f0a5a2c50806f0d5ab99dfd259ff3878","s_o_u_r_c_e__b_r_a_n_c_h__n_a_m_e":"feature/transactional-read-models","s_o_u_r_c_e__b_r_a_n_c_h__s_h_a":"d5c03dca7b2688f914eec50382b064d63ac9caad","t_a_r_g_e_t__b_r_a_n_c_h__n_a_m_e":"dev","t_a_r_g_e_t__b_r_a_n_c_h__s_h_a":"8a09550de60c4c9fee24708f959f76ad6f144156"}}

```

**Possible fixes**
Maybe error in https://github.com/jdalrymple/node-gitlab/blob/next/src/infrastructure/RequestHelper.ts#L21

[...]

PR #389, opened 2019-07-10
fix: Accept any variables that begin with uppercase letters and inclu…

jdalrymple:
…de an underscore #254

This updated regex isn't foolproof. If the user decides to use lowercase variables, this will not work. Would beed to change the internal logic to allow for the skipping of the camelization.

closes: #254

### DIFF
--- a/src/infrastructure/Utils.ts
+++ b/src/infrastructure/Utils.ts
@@ -1,4 +1,4 @@
-import { Constructor, Bundle } from ".";
+import { Constructor, Bundle } from '.';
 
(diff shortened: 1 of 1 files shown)


jdalrymple:
Im a little unsure as to why the tests are failing :/

jdalrymple:
```
 npm ci 
npm WARN prepare removing existing node_modules/ before installation
npm ERR! code MODULE_NOT_FOUND
npm ERR! Cannot find module './fetchers/version'
```
😕

PR #415, opened 2019-08-17
fix: incorrect url path for mr unapprove

talbright:
https://docs.gitlab.com/ee/api/merge_request_approvals.html#unapprove-merge-request


jdalrymple:
Oops! Thanks!

Issue #384, opened 2019-07-08
Documentation: Simple Typescript usage example


alaincao:
**Description**
Heylow, I'd like to have a simple example on where I can find the types returned by the different functions. e.g.
```typescript
import * as gitlab from 'gitlab'
const aaa = new gitlab.Gitlab({ host:'foo', token:'bar' });
const bbb : WHAT_AM_I??? = aaa;
```
The thing is that I would like to pass them to e.g. a function, and for now I am using `any` where the `WHAT_AM_I???` is, but that kinda defeats the purpose ...  
I am hitting a wall for *days*, here ; I have a hint with the
```typescript
import * as APIServices from 'gitlab/dist/services';
```
I saw in issue # 257, but still, no idea how to use ...

**Proposal**
I'd love to find a simple example e.g. in the `Readme.md` file or somewhere easily reachable ...

**Thanks !**

jdalrymple:
Ill throw something together in the next couple days!

jdalrymple:
So I haven't been able to get around to this. If anyone has time this would be a good first issue!

ksaaskil:
I was able to infer the type of `new Gitlab(...)` with following:
```ts
type ConstructedType<Constructor> = Constructor extends {
  new (...args: any[]): infer B;
}
  ? B
  : never;

type GitlabType = ConstructedType<typeof Gitlab>;
```
If that looks good, I can add a PR to add that in documentation and/or export the types for easier access?

ksaaskil:
No wait I'm dumb, one can use the built-in `InstanceType`:
```ts
type GitlabApi = InstanceType<typeof Gitlab>;
```

Bauke:
Although probably a rather big job, would it be possible to also get proper types for the responses to API calls?

For example calling `Projects.show()` returns a `Promise<GetResponse>`. If I then wanted to get the project ID or anything else that does in fact exist on the result, TypeScript errors out and says it doesn't exist on the `GetResponse` type (which would be correct):

```
source/edit-label.ts:24:48 - error TS2339: Property 'id' does not exist on type 'GetResponse'.
  Property 'id' does not exist on type 'object'.

24     log((await api.Projects.show(projectName)).id);
```

 Only by casting it to a type you made yourself with the right properties or `any` can you then access it:

```ts
    log((await api.Projects.show(projectName) as any).id);
    // Doesn't error and lets you access it
```

jdalrymple:
[...]

PR #425, opened 2019-08-27
Add type exports

ksaaskil:
- Exports the types of `Gitlab`, `UsersBundle`, etc. using the built-in `InstanceType` defined as follows:
```typescript
/**
 * Obtain the return type of a constructor function type
 */
type InstanceType<T extends new (...args: any) => any> = T extends new (...args: any) => infer R ? R : any;
```
- Should resolve the missing type problem of https://github.com/jdalrymple/node-gitlab/issues/384
```typescript
import * as gitlab from 'gitlab'
const aaa = new gitlab.Gitlab({ host:'foo', token:'bar' });
const bbb : WHAT_AM_I??? = aaa;
```
- Not sure if this could break anything, just a suggestion ¯\_(ツ)_/¯

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -87,3 +87,9 @@ export const ProjectsBundle = bundler({
 
 // All initialized
 export const Gitlab = bundler(APIServices);
+
+// Types
+export type UsersBundle = InstanceType<typeof UsersBundle>;
+export type GroupsBundle = InstanceType<typeof GroupsBundle>;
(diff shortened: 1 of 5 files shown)


casz:
Shouldn't import be `import { Gitlab } from 'gitlab'`

ksaaskil:
> Shouldn't import be `import { Gitlab } from 'gitlab'`

That would definitely work, I just copy-pasted the code from issue #384.

PR #448, opened 2019-09-27
Typedoc Generated Markdown

arsdehnel:
Added in `typedoc` and the `typedoc-plugin-markdown` to generate vuepress-friendly files.  I split it into two commits so you can see the `package.json` changes (in https://github.com/jdalrymple/node-gitlab/commit/f4ee64437267efae42fe9060ad31d735437e4958) and then the generated files (in https://github.com/jdalrymple/node-gitlab/commit/e2e5b1d7a31c66fd88db820aaaaba30cae68f378).


casz:
not sure I see much value in the generated docs same things my editor or IDE would tell me about the type.

Also due to the base service and resource interfaces it becomes a lot of repeated words, it sure is a nice way to get lots of additions 😆

jdalrymple:
@casz but unless you dl the library you cant see those docs :(. At keast this way people have docs before they install the libs

arsdehnel:
I’d rather have generated docs that have everything and are easy to keep up-to-date than a) what we have now or b) more complete docs that fall out-of-date or require a lot of work.

I think starting with a baseline where we generate what we can and then we can add more ‘curated’ docs as we mature it.

jdalrymple:
This will be much better once #1104 is finished!

Issue #302, opened 2019-04-11
The requested module 'gitlab' does not provide an export named 'Gitlab'


aol-nnov:
```
npm init
npm install --save gitlab@next
```

cat index.mjs

```
import { Gitlab } from 'gitlab';
```

```
node --experimental-modules index.mjs 
SyntaxError: The requested module 'gitlab' does not provide an export named 'Gitlab'
    at ModuleJob._instantiate (internal/modules/esm/module_job.js:80:21)
```

```
node -v
v10.6.0
```

but documentation on the `next` branch reads as 
```
// ES6 (>=node 10.0.0)
import { Gitlab } from 'gitlab';
```

What am I missing?

jdalrymple:
Thats a problem with the experimental mjs support. It doesn't seem to support named exports.  If you do:

```
import GitlabAPI from 'gitlab'

let {Gitlab} = GitlabAPI
```

aol-nnov:
@jdalrymple , thanks for the hints!

Also, I'd like to mention, that the following docs snippet from `next`

```
// ES5, assuming native or polyfilled Promise is available
const Gitlab = require('gitlab')


// Instantiating
const api = new Gitlab({
```

gives `TypeError: Gitlab is not a constructor` too

jdalrymple:
@aol-nnov Fixed! Sorry about that.

mariooc:
help, 
not work
try npm install --save gitlab@next
![gitlab-node](https://user-images.githubusercontent.com/2160877/56663730-250c0280-667d-11e9-96d2-24864f4dcfa5.png)

and  npm install --save gitlab
![gitlab-2](https://user-images.githubusercontent.com/2160877/56663959-9c419680-667d-11e9-8177-352099fea4ff.png)

jdalrymple:
Hmm, I think you can't install the package in a module with the same name.

casdevs:
Getting back to this...
I'm using gitlab@11.1.2 with a node.js v12 project and `--experimental-modules` switched on.

```import {Gitlab} from 'gitlab';``` gives me ```The requested module 'gitlab' does not provide an export named 'Gitlab'```.

However, when I remove ```"type": "module"``` from your module's package.json file, everything works as expected.

This statement was introduced somewhere between 11.0.0 (here, everything works as expected) and 11.1.2.

[...]

PR #453, opened 2019-10-01
fix: removed "type": "module" from package.json

casdevs:
because main export still uses CommonJS modules, not ES6 import/export

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,6 +101,5 @@
     "test:integration": "jest test/integration -c=jest.config.json --detectOpenHandles",
     "test:unit": "jest test/unit --debug -c=jest.config.json"
   },
-  "type": "module",
   "types": "dist/index.d.ts"
 }


casdevs:
this should solve #302 or at least fix import problems when using this package in node projects with --experimental-modules enabled and "type": "module" in their package.json file

jdalrymple:
I could just be missing something, but the main export is using the export syntax? https://github.com/jdalrymple/node-gitlab/blob/master/src/core/index.ts#L90

casdevs:
ya sure, but it is relevant what you're exporting into the final NPM package after rollup has finished its transpiling.

to cite the node.js docs:

> Files ending with .js or .mjs, or lacking any extension, will be loaded as ES modules when the nearest parent package.json file contains a top-level field "type" with a value of "module".

see https://nodejs.org/api/esm.html#esm_code_package_json_code_code_type_code_field

your "main" export `index.js` clearly is ES5/CommonJS format, so it doesn't seem correct to tell node.js that it is an ES module (which only triggers error in recent node.js versions and when using the `--experimental-modules` flag).
ES module users should be fine anyway, as you also export an ES module "index.es.js" and tell node about it via `"module": "dist/index.es.js"`

jdalrymple:
You're 100% correct! Totally forgot about that for a second! Thanks for the PR!!

casdevs:
great, thanks for merging that in!

PR #461, opened 2019-10-09
Fix: epic issues http methods were wrong

arsdehnel:
Not sure why these were `put/del` instead of `post/put` but these aren't working right now, this should fix that problem.

### DIFF
--- a/src/core/services/EpicIssues.ts
+++ b/src/core/services/EpicIssues.ts
@@ -17,13 +17,13 @@ class EpicIssues extends BaseService {
   assign(groupId: GroupId, epicId: EpicId, issueId: IssueId, options?: Sudo) {
     const [gId, eId, iId] = [groupId, epicId, issueId].map(encodeURIComponent);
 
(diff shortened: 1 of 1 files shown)


jdalrymple:
I probably read the wrong endpoint smh. Thanks!

arsdehnel:
Do you know when this might get tagged?  I've got a project I'm working on and would like to just keep using this one rather than having to go through and publish my own fork to our private npm registry and stuff.

jdalrymple:
Didnt notice the missing keyword from the commit message. Ill fix this today :)

arsdehnel:
My bad.  What's the keyword I need to include?  Maybe I missed something in the docs?

jdalrymple:
It's not your fault :) I dont mentioned it anywhere, it has to do with semantic release. To trigger the builds, each commit must be prefixed with the type of change it was. So for this example, "fix: Epic issues http methods were wrong", which is what you named the PR (which is good) but it needs to be on the commit to trigger the build.

jdalrymple:
Should be released soon :)

Issue #428, opened 2019-08-28
Somehow figure out when the GL instance is 100% up and running before running tests


jdalrymple:
**Description**

Currently the tests wait a specified amount of time before beginning after starting up the gitlab instance. This is needed because a key needs to be generated before any of those tests can run. To create this token, we make an API request to the GL instance and if the instance isnt up, the token remains null and subsequently fails all the tests. 

Not only is this inefficient but prone to error if the instance isn't up in that allotted time. In the past we tried to figure this out based on the docker up time endpoints ([liveliness and readiness](https://docs.gitlab.com/ee/user/admin_area/monitoring/health_check.html)) but they didnt help, at least not a year ago. I would love to fix this and avoid erroneous failed builds :upside_down_face: 

**Steps to reproduce**
Run the integration tests, but have the execute without a time delay. 

**Expected behaviour**
Tests wait until GL instance is ready before running

**Actual behaviour**
Using a fixed time delay before executing

**Possible fixes**
Maybe readiness works now? Or perhaps changing the Gitlab runner container such that these endpoints actually work as they should?

casz:
Maybe we can create a specific docker container for it?

You could add a simple nodejs container sidecar that you repeat attempting to login unless you get 200?

Perhaps look into: https://github.com/testcontainers/testcontainers-node

casz:
For your testing I would properly switch to https://github.com/testcontainers/testcontainers-node it can even do await successful response

casz:
Ya but you want a second container or some code that waits.
Does a login and starts navigating to health_check or the first project to ensure that the gitlab container is ready.

So wait till port `80` and `433` respond then attempt login and keep repeating till you get logged in.
Then head to first project and keep trying till get a `200` respond.
You could go further and have code that attempts to create a issue and wait till it successfully manages to.

casz:
Well one will help prove the second? 😆

casz:
That could work too :)

PR #464, opened 2019-10-10
test: Add a readiness test to avoid improper integration test failure

jdalrymple:
closes: #428


test/docker/test_readiness.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+check_readiness() {
```
casz: Suggestion, write the script in ruby or node.js?

casz:
response not a command 😆

casz:
You still need to prefix variables when referenced afterwards with `$` so

```bash
local db_check=$($response | jq '.db_check.status')
```

test/docker/test_readiness.sh
jdalrymple: I could! I was just trying to keep it as simple of a script as possible

test/docker/test_readiness.sh
```diff
@@ -0,0 +1,22 @@
...
+   local response=$(curl -s "${GITLAB_URL}")
```
casz: Shouldn't the URL be `/-/readiness`?

jdalrymple: Yea, I just put some template url there for now as i debug my syntax errors

test/docker/test_readiness.sh
```diff
@@ -0,0 +1,28 @@
...
+   local response=$(curl -s "${GITLAB_URL}/-/readiness")
+
+   local db_check=$(echo "$response" | jq '.db_check.status')
+   local redis_check=$(echo "$response" | jq '.redis_check.status')
+   local cache_check=$(echo "$response" | jq '.cache_check.status')
+   local queues_check=$(echo "$response" | jq '.queues_check.status')
+   local shared_state_check=$(echo "$response" | jq '.shared_state_check.status')
+   local gitaly_check=$(echo "$response" | jq '.gitaly_check.status')
+
+   return db_check == "ok" && redis_check == "ok"  && cache_check == "ok" && queues_check == "ok"  && shared_state_check == "ok" && gitaly_check == "ok"
```
casz: You still need to prefix these guys 🗡 
```suggestion
   return $db_check == "ok" && $redis_check == "ok"  && $cache_check == "ok" && $queues_check == "ok"  && $shared_state_check == "ok" && $gitaly_check == "ok" 
```

casz: Most likely you will need to write something like this

```suggestion
   if [ "$db_check" == "ok" && "$redis_check" == "ok" && "$cache_check" == "ok" && "$queues_check' == "ok" && "$shared_state_check" == "ok" && "$gitaly_check" == "ok"  ]; then
     true
   else
     false
   fi
```

casz:
You can wrap them all in one `[]` instead of individually 😓 

You properly need to wrap the while loop too

casz:
Suggestion you can use docker image built in health-check and docker ps to determine when the container is ready.

https://gitlab.com/gitlab-org/omnibus-gitlab/blob/master/docker/Dockerfile#L46-47

```
docker ps --filter "health=healthy" -q
```
Then just wait till you get a container ID back

jdalrymple:
Im not sure if that docker container check will actually return a correct status for the readiness of the gitlab instance

jdalrymple:
Bash will be the end of me lol

casz:
Should be simple to write in nodejs with no dependencies? 😆

jdalrymple:
At this point perhaps :/

casz:
My experience over at https://github.com/nmklotas/GitLabApiClient/pull/68 tells me that the health check will report by when the gitlab is "healthy" which means that gitlab is ready and takes roughly the same time for GitLab to startup 2-3 minutes depending on the machine it is running on.

jdalrymple:
I must be missing something. When querying the instance for readiness, i get a "Not found" error. The url im hitting is: http://localhost:8080/-/readiness, which is the same base url i make to perform the API calls (http://localhost:8080/api/v4/...)

jdalrymple:
Alright, I think my not found error has to do with ip whitelisting. Looking into that

casz:
Hmm I might have to copy how you spin up Travis, cause I cannot get my test instance to get pass starting the postgres database 😭

Issue #310, opened 2019-05-08
Support graphql feature


casz:
**Description**
Support GraphQL API queries

**Proposal**
Would be great to already start implementing the use case for using Gitlab's [new GraphQL API](https://docs.gitlab.com/ee/api/graphql/)

casz:
Updated with an example query :sweat_smile:

jdalrymple:
Would we just expose a GraphQL model which would be initialized with the users settings, and then it would have one function "query" which would take in a graphql query like above?

casz:
I think that would be wonderful :sweat_smile:

casz:
The query should of course support variables as well :sweat:

jdalrymple:
Not sure exactly how it would all work, but i guess we could brain storm some things here!

casz:
The easiest thing would be pull in a GraphQL client? No need to rewrite everything.
Though not that it requires that much code: https://www.npmjs.com/package/graphql-request
https://github.com/prisma/graphql-request/blob/master/src/index.ts

jdalrymple:
Keeping things simple, I like it. Easy to include the graphQL client. This could probably be done relatively quickly...

jdalrymple:
I was thinking about this today. What would be the best way to expose this? As a service? so you'd import it like:

```
import { GitLabQL } from 'gitlab';
```

jdalrymple:
Would this be something only available to node users? As in though this be its own package @gitbeaker/graphql ?

jetersen:
This would be beneficial to both node and browser :)

jdalrymple:
Can you make requests to the graphql api through standard http requests? or does it require a graphql requester that supports browser and node?

jetersen:
you can do with normal http requests it is just post method.
Though there is some special logic handling queries.

jdalrymple:
> you can do with normal http requests it is just post method.
> Though there is some special logic handling queries and pagination.

Any examples of that special logic. I could try to incorporate it into the Got/Ky Requesters

jdalrymple:
Rereading the docs, this is  very doable, the only issue is determining how the pagination would work :thinking: Would there be any? Or would the user need to handle that themselves

jdalrymple:
I definitely want to get to this, just need to find the time :disappointed:

ThePlenkov:
@jdalrymple  i can imagine - I try now to create generated typescript types out of gitlab graph ql scema.

[...]

PR #466, opened 2019-10-10
Add GraphQL Support

jdalrymple:
Still need to test, and cleanup. This is just so I don't forget.

#Questions 
- How would this endpoint play out for the cli support?
- Should we support variables off the bat?
- Would this work fine in the browser?

closes: #310


jdalrymple:
Closed to open a draft :s

PR #470, opened 2019-10-12
fix spelling and grammer error

SRAVANAAJ:


jdalrymple:
Thanks! :man_facepalming:

PR #477, opened 2019-10-21
fix: resolve import issues

DanielRuf:
This PR resolves the issues with the wrong imports, as these are not modules but cjs.


src/core/infrastructure/KyRequester.ts
```diff
@@ -26,7 +25,7 @@ function defaultRequest(service: any, { body, query, sudo, method }) {
   if (sudo) headers.append('sudo', `${sudo}`);
 
   if (typeof body === 'object' && !(body instanceof FormData)) {
-    bod = JSON.stringify(decamelizeKeys(body, skipAllCaps));
+    bod = JSON.stringify(decamelizeKeys(body));
```
jdalrymple: Does it avoid decamelizing if the input is in the form of "SOME_UPPER_CASE_VARIABLE" ?

DanielRuf: Screenshot_20191022-050808

jdalrymple:
@DanielRuf any suggestions for an xcase replacement? Though it says it supports "some-string" to camelcase, it doesn't seem to function properly https://runkit.com/embed/tlx14hnn6mux

PR #493, opened 2019-11-13
add downloadSingleArtifactFileFromRef method

tiger-seo:
Download a single artifact file from specific tag or branch
https://docs.gitlab.com/ee/api/jobs.html#download-a-single-artifact-file-from-specific-tag-or-branch


jdalrymple:
LGTM!

PR #494, opened 2019-11-14
feat: Add Group.projects

Granjow:
This commit also introduces a minimal typed response as a first step
towards #384.

### DIFF
--- a/src/core/services/Groups.ts
+++ b/src/core/services/Groups.ts
@@ -5,6 +5,7 @@ import {
   RequestHelper,
   Sudo,
 } from '../infrastructure';
+import { ProjectSchema } from './Projects';
 
 export class Groups extends BaseService {
   all(options?: PaginatedRequestOptions) {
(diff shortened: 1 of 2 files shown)


jdalrymple:
The only suggestion id have would be moving the interface information into the corresponding service, just for consistency.

Granjow:
@jdalrymple Yes, I did not do this because also other endpoints return the same structure, e.g. /api/v4/projects returns projects as well. But I could move it to servcies/Projects.ts and reference it from Groups.ts?

Granjow:
Anyone, can this MR be merged?

jdalrymple:
Yea sorry! I though i responded :man_facepalming: I would move it into the Projects.js file!

Granjow:
@jdalrymple Updated! I suffixed the interface with `Schema` so people do not accidentially import `Project` instead of `Projects`.

PR #502, opened 2019-11-27
fix: Adding back functionality for rejectUnauthorized http option

jdalrymple:


jdalrymple:
Latest change should handle this: https://github.com/node-gitlab/node-gitlab/issues/357#issuecomment-567528152

@casz give it a quick look :pray:

casz:
conflicts in package.json 😱

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "gitlab",
   "description": "Full NodeJS implementation of the GitLab API. Supports Promises, Async/Await.",
-  "version": "14.0.0",
+  "version": "14.0.1",
```
casz: seems odd that this is touching version 🤔

jdalrymple: Just because i was pushing to a beta tag on npm

Issue #507, opened 2019-12-09
Add support for all inherited members endpoint (ancestor groups) - /groups/:id/members/all/:user_id


MadLittleMods:
Add support for the following endpoints as documented in https://docs.gitlab.com/ee/api/members.html#get-a-member-of-a-group-or-project-including-inherited-members

 - **`GET /groups/:id/members/all/:user_id`**
 - **`GET /projects/:id/members/all/:user_id`**

This endpoint is special because it "includ[es] members inherited through ancestor groups."

We are using this in Gitter to support GitLab based communities/rooms, https://gitlab.com/gitlab-org/gitter/webapp/merge_requests/1679

---

Normal endpoints which don't include inherited ancestor group members (these are already supported), https://docs.gitlab.com/ee/api/members.html#get-a-member-of-a-group-or-project

```
GET /groups/:id/members/:user_id
GET /projects/:id/members/:user_id
```

casz:
this is already supported.
https://github.com/node-gitlab/node-gitlab/blob/63b11574aa1db3b88be3b3baeb16dbaceb9e566b/src/core/templates/ResourceMembers.ts#L17

set `includeInherited` to true

casz:
ah, this was introduced in `GitLab 12.4.` as note by https://docs.gitlab.com/ee/api/members.html#get-a-member-of-a-group-or-project-including-inherited-members

So obviously this is currently not supported 😓

casz:
btw I love the fact that Gitlab/gitter is using this library 🥰

MadLittleMods:
@jdalrymple @casz I got this [working and all integrated](https://gitlab.com/gitlab-org/gitter/webapp/merge_requests/1679) now 🚀 Thanks for the great support and quick update  ❤️

jdalrymple:
@casz Next step is to clean it up a bit #483 ;)

PR #508, opened 2019-12-09
Resource Members show to include inherited members

jdalrymple:
feat: Adding ability to get a singular member of a project or group, including inheritedMemebrs

BREAKING CHANGE: Updated the Members.all function to have the inheritedMembers option to be in the optional object.

closes: #507

### DIFF
--- a/src/core/templates/ResourceMembers.ts
+++ b/src/core/templates/ResourceMembers.ts
@@ -6,15 +6,21 @@ import {
   RequestHelper,
   Sudo,
 } from '../infrastructure';
-
 import { AccessLevel } from './ResourceAccessRequests';
+
+interface IncludeInherited {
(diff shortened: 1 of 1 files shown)


src/core/templates/ResourceMembers.ts
```diff
@@ -6,15 +6,21 @@ import {
   RequestHelper,
   Sudo,
 } from '../infrastructure';
-
 import { AccessLevel } from './ResourceAccessRequests';
 
+interface IncludeInherited {
+  includeInherited?: boolean;
```
casz: hmmm

Argument of type '{ includeInherited: boolean; }' is not assignable to parameter of type 'boolean'.ts(2345)

`api.ProjectMembers.all(project.id, { includeInherited: false })`

Issue #483, opened 2019-10-24
Updating Group Organization - Thoughts?


jdalrymple:
**Description**
Just thinking about how to better organize the group for ease of use by developers looking for Gitlab add-ons.

1. Similar to the thoughts in https://github.com/moul/node-gitlab/issues/191#issuecomment-381260105 It would be nice to clean up the group. Maybe change it to some generic name to host gitlab wrappers? Where each internal repo would be node-gitlab, python-gitlab etc
2. Archiving older repos in the group that are no longer in use or are duplicates?
3. Mirroring on gitlab. Right now its just on my personal account, but since weve moved the repo to this group, it has become inconsistent.
4. Keep things as they were, redirect traffic from the archived repo

Anyways just looking for thoughts!! :D

casz:
I see you already moved it to node-gitlab/node-gitlab 😆

jdalrymple:
lol, we can always move it around :P hence the request for Ideas. One of the biggest issues was that the google searches were pointing to node-gitlab/node-gitlab which isnt really useful for those looking for this repo.

jdalrymple:
*cricket sounds* Kinda leaning towards #4 since it would be a bit less work lol

casz:
I think you could split the bin out into a separate repo 😄
To avoid unnecessary dependencies for those who do not use the bin?

jdalrymple:
We could also do a monorepo and split the packages similarly to jest

casz:
Ya that could work too lerna is awesome 😄

jdalrymple:
I agree!! But that still doesnt solve what we should do about the group lol

casz:
Still not sure what you had in mind for the group, do you want to move all gitlab wrappers to a new group? ie. 
Do you want to move something like https://github.com/nmklotas/GitLabApiClient ?

jdalrymple:
That's what I'm trying to figure out, whats the purpose of the group?

jdalrymple:
Side note: We could setup an organization, and publish for cli, core (nodejs) and move the types to the @types repo or its own package. This would reduce the size of the published packages as well. 

Problems with this move are the change of the npm package name :/ and intuitive organization name (@gitlab is already taken)

Just thinking outloud! More ideas are welcome :D

jdalrymple:
So I've come to a decision that I'm going to try and complete over the holidays. Basically, I want to follow along the octokit path and rebrand the repo/npm package.

[...]

PR #517, opened 2019-12-22
feat: vulnerability findings API support

arsdehnel:
Adding support for fetching all findings: https://docs.gitlab.com/ee/api/vulnerability_findings.html#list-project-vulnerability-findings

### DIFF
--- a/src/core/services/VulnerabilityFindings.ts
+++ b/src/core/services/VulnerabilityFindings.ts
@@ -1,0 +1,9 @@
+import { BaseService, PaginatedRequestOptions, RequestHelper } from '../infrastructure';
+
+export class VulnerabilityFindings extends BaseService {
+  all(projectId: string | number, options?: PaginatedRequestOptions) {
+    const pId = encodeURIComponent(projectId);
+
+    return RequestHelper.get(this, `projects/${pId}/vulnerability_findings`, options);
+  }
+}
--- a/src/core/services/VulnerabilityFindings.ts
+++ b/src/core/services/VulnerabilityFindings.ts
@@ -5,11 +5,11 @@ export class VulnerabilityFindings extends BaseService {
     projectId: string | number,
     options?: {
       id: string | number;
-      report_type: string[];
+      reportType: string[];
       scope: string;
       severity: string[];
       confidence: string[];
-      pipeline_id: string | number;
+      pipelineId: string | number;
     } & PaginatedRequestOptions,
   ) {
     const pId = encodeURIComponent(projectId);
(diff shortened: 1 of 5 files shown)


src/core/services/VulnerabilityFindings.ts
```diff
@@ -0,0 +1,9 @@
+import { BaseService, PaginatedRequestOptions, RequestHelper } from '../infrastructure';
+
+export class VulnerabilityFindings extends BaseService {
+  all(projectId: string | number, options?: PaginatedRequestOptions) {
```
casz: Perhaps add the types for the query options?

arsdehnel: Updated with what I think you were after.  I haven't done anything with Typescript before so this is all a bit new, let me know if that's not what you were thinking.

Having said that I'm a noob on TS I did wonder if making things like the `severity`, `confidence` and `report_type` into enums would make sense here?  I didn't see any enums anywhere in the project so wasn't sure if there was a reason to avoid them or not.

casz: in ts, enum is as simple as saying `'undefined' | 'ignore'` etc :)

enums do exist in typescript: https://www.typescriptlang.org/docs/handbook/enums.html but in this case I think possible values are faster.

casz: otherwise LGTM

jdalrymple: Nice addition! I would suggest making the properties camelCase for consistency. Other than that LGTM

arsdehnel: I’ll change to camelCase, that was a result of me being consistent with the gitlab API, my bad.

casz: we got the library that takes care of the casing 😉

arsdehnel: updated with camelCase

arsdehnel:
I'm done messing with this one unless there are more items you'd like adjusted.

jdalrymple:
LGTM, im just cleaning up the repo. Ill merge/ release soon! Thank you!

arsdehnel:
Is that "cleaning up" why I see this showing up as `jdalrymple/gitbeaker` now?

jdalrymple:
> Is that "cleaning up" why I see this showing up as `jdalrymple/gitbeaker` now?

Yup! See #483 for more info! I moved everything into a separate branch so i should be able to do things in parallel ;)

PR #521, opened 2019-12-29
fix: make ResourceMembers.all/show delivery correct options params

### LINKED ISSUE #518: pagination does not work for ResourceMembers
"gitlab": "^14.0.1",

```js
// Jest
test('api /projects/${repoName}/members', async function (done) {
    let members = await api.ProjectMembers.all('namespace/name', {
      page: 1,
      showPagination: true,
      maxPages: 3
    }) as any;
    expect(Array.isArray(members)).toBe(true);  // always be true here
    console.log(members)
    done();
 });
```
when I use `ProjectMembers` to get members with `PaginatedRequestOptions`,  it always return array. 

I tried `ProjectMembers` and some other services, they work fine.

I got the compiled position in dist/index.js （`https://github.com/jdalrymple/gitbeaker/blob/c6f4c51d585c3906547334256e01c31e1ddced14/src/core/infrastructure/RequestHelper.ts#L78`） ，write  'console.log' here to see what happened, the log makes me confused.

dist/index.js
```
case 3:console.log(p);return[2,(p.page||c.length>0)&&o?{data:c,pagination:a}:c]
```

terminal output
```
  ✓ api /projects/${repoName}/members (203ms)

  console.log node_modules/gitlab/dist/index.js:1
    {}

  console.log node_modules/gitlab/dist/index.js:1
    { options: { page: 1, showPagination: true, maxPages: 3 } }
```

Shouldn't it only be executed once，and the console.log(p) print `{ page: 1, showPagination: true, maxPages: 3 }` ?

jiadi0801:
fix #518

### DIFF
--- a/src/core/templates/ResourceMembers.ts
+++ b/src/core/templates/ResourceMembers.ts
@@ -70,7 +70,7 @@ export class ResourceMembers extends BaseService {
 
     url.push(uId);
 
-    return RequestHelper.get(this, url.join('/'), { ...options });
+    return RequestHelper.get(this, url.join('/'), options);
   }
 
   remove(resourceId: string | number, userId: number, options?: Sudo) {
--- a/src/core/templates/ResourceMembers.ts
+++ b/src/core/templates/ResourceMembers.ts
@@ -26,7 +26,7 @@ export class ResourceMembers extends BaseService {
 
     if (includeInherited) url.push('all');
 
-    return RequestHelper.get(this, url.join('/'), { options });
+    return RequestHelper.get(this, url.join('/'), { ...options });
   }
 
   add(
@@ -70,7 +70,7 @@ export class ResourceMembers extends BaseService {
 
     url.push(uId);
 
-    return RequestHelper.get(this, url.join('/'), { options });
+    return RequestHelper.get(this, url.join('/'), { ...options });
   }
 
   remove(resourceId: string | number, userId: number, options?: Sudo) {
--- a/src/core/templates/ResourceMembers.ts
+++ b/src/core/templates/ResourceMembers.ts
@@ -26,7 +26,7 @@ export class ResourceMembers extends BaseService {
 
     if (includeInherited) url.push('all');
 
-    return RequestHelper.get(this, url.join('/'), { ...options });
+    return RequestHelper.get(this, url.join('/'), options);
   }
 
   add(


src/core/templates/ResourceMembers.ts
```diff
@@ -26,7 +26,7 @@ export class ResourceMembers extends BaseService {
 
     if (includeInherited) url.push('all');
 
-    return RequestHelper.get(this, url.join('/'), { options });
+    return RequestHelper.get(this, url.join('/'), { ...options });
```
casz: instead of unpacking, I should we just toss the `options` object at it.
```suggestion
    return RequestHelper.get(this, url.join('/'), options);
```

src/core/templates/ResourceMembers.ts
```diff
@@ -70,7 +70,7 @@ export class ResourceMembers extends BaseService {
 
     url.push(uId);
 
-    return RequestHelper.get(this, url.join('/'), { options });
+    return RequestHelper.get(this, url.join('/'), { ...options });
```
casz: instead of unpacking, I should we just toss the `options` object at it.
```suggestion
    return RequestHelper.get(this, url.join('/'), options);
```

jiadi0801:
> Hmm I wonder if a quick search will reveal more of these?

yeah, I use RequestHelper to search in /src,  found that addNote/editNote methods in `ResourceDiscussions.ts` used wrong params to invoke RequestHelper.put/post (https://docs.gitlab.com/ee/api/discussions.html#add-note-to-existing-issue-thread)。the `discussionId` in gitlab api v4 is a hash value, so the type declaration may be `number | string`.

 Should I fix it in this PR or create a new issue(another PR)?

jdalrymple:
Thank you!!

jdalrymple:
> > Hmm I wonder if a quick search will reveal more of these?
> 
> yeah, I use RequestHelper to search in /src, found that addNote/editNote methods in `ResourceDiscussions.ts` used wrong params to invoke RequestHelper.put/post (https://docs.gitlab.com/ee/api/discussions.html#add-note-to-existing-issue-thread). The `discussionId` in gitlab api v4 is a hash value, so the type declaration may be `number | string`.
> 
> Should I fix it in this PR or create a new issue(another PR)?

Perhaps another PR, better to have small PRs.

PR #533, opened 2020-01-12
Gitbeaker - Offical Release 🚀

jdalrymple:
# Release
## Core
**chore**

- Cleaning up the eslint configuration
- Cleaning up the lintstaged configuration
- Updated the codecov configuration to handle the monorepo structure
- Added a global.d.ts file to solve the jest-extended typing errors

**docs**

- Updated License year
- Added logo assets and updated the README

**refactor**

- Switched to yarn and used lerna to implement a monorepo structure
- Removed creation of temp files during the api map generation
- Split the Requester logic to handle both browser and node to their fullest abilities. The node release makes use of the Got library (with streaming), where as the browser release makes use of ky.

**test**

- Added tests for the GotRequester
- Moved CI to gitlab-ci **BUT** currently have the integration tests commented out until the gitbeaker-test branch is fixed.
- Switched the testing framework to provide coverage only on the tests that point to the src. Further tests have been added to target the dist versions.

**fixes**

- Fixed notification settings to handle their odd put request calls
- Added missing camelized versions of the schema types
- ResourceTemplate was using the incorrect http method for showing a single template

**BREAKING**

- Group creation now takes three parameters since the path property was required but not included in the non-optional parameters.
- Changed the spelling of the License API from `Licence` to `License`
- license.create method has been replaced with license.add and has a required first parameter that is the license
- license.all method returns all licenses and no longer the license for the current user. This is now done by license.show
- ResourceAwardEmojis: The order of arguments that include both the awardId and noteId has changed. It is now projectId, resourceId, noteId, and then awardId. This affects the award, remove and show functions.

## CLI

**feat**
- Added support for GITBEAKER_[Property] env variables as well as the `gb-[property]` prefix


packages/cli/src/index.ts
```diff
@@ -20,58 +18,75 @@ const hintStyle = chalk.hex('#6a5f88');
 
 // Globally configurable arguments
 const globalConfig: { [name: string]: Options } = {
-  'gl-token': {
+  'gb-token': {
+    alias: 'gl-token',
     desc: 'Your Gitlab Personal Token',
     type: 'string',
-    defaultValue: process.env.GITLAB_TOKEN,
+    defaultValue: process.env.GITBEAKER_TOKEN || process.env.GITLAB_TOKEN,
```
casz: First big uff.
I think this is going beyond the branding. This clearly is a GITLAB_TOKEN and this is something gitlab ci will even expose.

jdalrymple: Yea there is gonna be alot of clean ups on the way as I move it all to the monorepo. That's why I'm stressing the test coverage. The change had it previously , i just added the gitbeaker part. You think keeping the GITLAB prefix might cause some conflicts down the road?

casz: I must admit I missed the `||` and only saw `GITBEAKER_TOKEN`.

Fine as long as `GITLAB_TOKEN` stays. I see no conflicts I was worried that we only had `GITBEAKER_TOKEN` which would IMHO break a lot of things.

jdalrymple: I agree, thats why i kept it :D Should all still work fine

jdalrymple:
So, I've got the coverage up, but I'm running into testing problems. 

First, occasionally the API requests inside the beforeAll functions don't fulfill correctly, leaving all the tests that depend on that beforeAll to fail. This is not consistent as can be seen [here](https://gitlab.com/jdalrymple/gitbeaker/-/jobs/403077138). I've rerun that test a few times and occasionally it passes perfectly, occasionally it fails due to a missing project.id which means project is undefined. 

Secondly, occasionally there is a 502 Bad Gateway error from the gitlab server. I'm guessing this happens if to many requests are made in a short period of time, so I've switched to have things runInBand. This is quite annoying and hasn't particularly solved the problem. I'm a little unsure of how to fix this one.

I've tried testing each test suite one by one, and they all pass on their own without either problem #1 or #2. So I'm unsure of how to handle this. Perhaps its a lingering promise? or Jest isnt handling the asyncs well? 

**Possible Solutions**
For the first problem, I've moved everything from the global beforeAlls into the actual tests. This makes things much dirtier, but should remove any problems with test dependencies. Unfortunately this means many more requests to the Gitlab server, so it may make problem #2 much worse.

Any other suggestions?

jdalrymple:
I've begun running these tests on a local docker image and have managed to recreate this inconsistent behavior. Hopefully I can narrow it down and report it to gitlab!

casz:
@jdalrymple see my approach in https://github.com/nmklotas/GitLabApiClient

Do not use persisted storage.
Create a group and project on startup.
Use custom gitlab.rb config file.
For those tests that create projects/groups I delete afterwards.

jdalrymple:
Hmm not sure that will fix the problem. Running it locally in a socket works flawlessly. But it still performs badly up in the ci. 🤷🏾‍♂️

casz:
👀

README.md
```diff
@@ -65,20 +97,20 @@ const api = new Gitlab({
...
+| `rejectUnauthorized` | Yes      | `true`                                                  | Http Certificate setting, Only applies to HTTPS hosts urls                                                         |
+| `sudo`               | Yes      | `false`                                                 | [Sudo](https://docs.gitlab.com/ee/api/#sudo) query parameter                                                       |
+| `version`            | Yes      | `4`                                                     | API Version ID                                                                                                     |
+| `camelize`           | Yes      | `false`                                                 | Camelizes all response body keys                                                                                   |
+| `requester`          | Yes      | [GotRequester.ts](./src/infrastructure/GotRequester.ts) | Request Library Wrapper. Currently wraps Ky.                                                                       |
```
casz: Which is it? Ky, Got, or both?
```suggestion
| `requester`          | Yes      | [GotRequester.ts](./src/infrastructure/GotRequester.ts) | Request Library Wrapper. Currently wraps Got.                                                                       |
```

lerna.json
```diff
@@ -0,0 +1,6 @@
+{
+  "packages": ["packages/*"],
+  "version": "14.0.1",
```
casz:
```suggestion
  "version": "14.2.1",
```

packages/core/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "name": "Justin Dalrymple"
+  },
+  "browser": "dist/index.umd.js",
+  "bugs": {
+    "url": "https://github.com/jdalrymple/gitbeaker/issues"
+  },
+  "dependencies": {
+    "form-data": "^3.0.0",
+    "got": "^10.2.2",
+    "ky": "^0.16.1",
```
casz: something smells if we need to depend on both 🤔

packages/core/src/infrastructure/KyRequester.ts
```diff
@@ -0,0 +1,66 @@
+import Ky from 'ky';
```
casz: Should this not be located in the browser portion?

README.md
jdalrymple: I totally missed this! Thank you

lerna.json
jdalrymple: I think this number will be bumped to 15, but im not sure :s

packages/core/package.json
jdalrymple: Yea, i couldnt think of the best way to split out the browser version. Its just a diff rollup config where i switch the deps from got to ky. If you have any suggestions id be open to trying something else?

packages/core/src/infrastructure/KyRequester.ts
jdalrymple: See previous comment

packages/core/package.json
jdalrymple: The only thing i could think of was having another package for the browser that used the core source with a custom Requester for Ky and its own rollup config. BUT that would require depending on the source of another project, which im not currently doing. Not sure if thats a good practice. Ill look up some examples is that done

casz: What about `@gitbeaker/core` with no dep to ky or got?
Than `@gitbeaker/node` with got and `@gitbeaker/browser` with ky?

casz: Then it even opens up to someone depends on `@gitbeaker/core` where they write their own requester.

jdalrymple: > What about `@gitbeaker/core` with no dep to ky or got?
> Than `@gitbeaker/node` with got and `@gitbeaker/browser` with ky?

This...this may be fine. Ill try it!

jdalrymple: This was a good idea! Thank you. Not sure if I should split out the infrastructure code into a helpers package because right now, the export for core is { all the services, all the bundles, and one helper class}. This requires deconstructuring if you want to import everything but not the helper. Not sure if its worth the hassle though.

It would also be nice to fix the path for the APIMap import which is now done though:

'import * as map from '@gitbeaker/core/dist/map.json'

Any ideas?

casz: Why not move the RequestHelper out to it's own package? x number of helpers?

You could also remove the RequestHelper from the `index.ts` and require that they import it directly.

jdalrymple: I split it off, its a little verbose, but more flexible

jdalrymple: As for the map thing, one suggestion is modifying the prepublish script to move dist files to root , and updating the files property of the package.json to include the few files i want exported. This would result in the export being:

```
'import * as map from '@gitbeaker/core/map.json'
```

Ill leave this for now, but something to think about

casz:
Looks really promising :)

jdalrymple:
I think we're good now. Give it a look over, find the problems I might of missed! :joy:

packages/gitbeaker-cli/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "name": "Justin Dalrymple"
+  },
+  "bin": {
+    "gitbeaker": "dist/index.js"
+  },
+  "bugs": {
+    "url": "https://github.com/jdalrymple/gitbeaker/issues"
+  },
+  "dependencies": {
+    "@gitbeaker/core": "15.0.0",
```
casz: This should be pulled in from `@gitbeaker/node`?

README.md
```diff
@@ -40,7 +70,12 @@
 
 ```bash
 # Install from npm
-npm install gitlab
+npm install @gitbeaker/node # NodeJS default, index.es.js for esm
```
casz: missing example of importing/require?

packages/gitbeaker-cli/package.json
jdalrymple: I need to import the map generated in the core module :( The good thing is that the rollup tree shaking will only pull the json file and not the whole package

README.md
jdalrymple: Ill add that!

packages/gitbeaker-cli/package.json
casz: You should still have access to the dep because `@gitbeaker/node` depends on it? 🤔

jdalrymple: Its its not included in the dependencies, I cant import it in the code though... Unless I'm missing something.

jdalrymple: Just waiting on thoughts for this^

casz: might be different with lerna so seems fine 😅

casz:
Looking forward to a release that I can consume 🙇
Great work! 🥇

jdalrymple:
Only took 100 years, but its been released!

Issue #574, opened 2020-01-25
Add detailed search support to Merge Requests


AlvaroBernalG:
Hello good people of gitbeaker,

**Description**

It would be great if we could search for Merge Requests inside a project by matching a text against MR title or/and description.

**Proposal**

Proposed API:

```js

const searchParams: MergeRequestSearchParams = {};

api.MergeRequests.search(searchParams);
```
MergeRequestSearchParams will contains all the available merge request params documented in: 
https://docs.gitlab.com/ee/api/merge_requests.html#list-merge-requests 
 plus another one: `projectId`. If `projectId` is present then the endpoint will change and point to: 
https://docs.gitlab.com/ee/api/merge_requests.html#list-project-merge-requests.

jdalrymple:
You can currently do this by using the .all method! :smile: . This should work like this:

```javascript
const results = await MergeRequests.all({projectId: 6, state:'all', labels:'bug'});
```

If the projectId is omitted, it will search all MR's. You can also use the groupId instead to get all MR's corresponding to a specific group

AlvaroBernalG:
I will create then a PR mapping the Gitlab endpoint params to a interface so that we can have IntelliSense: 

<img width="816" alt="Screenshot 2020-01-25 at 16 36 22" src="https://user-images.githubusercontent.com/12004383/73124239-fc5cc480-3f90-11ea-9908-487838502381.png">



Or is there a reason why the options object does not have an interface, like .create()? 

 <img width="639" alt="Screenshot 2020-01-25 at 16 43 27" src="https://user-images.githubusercontent.com/12004383/73124306-dab00d00-3f91-11ea-91db-dea762abd9e5.png">

jdalrymple:
No reason, just never had the time to add all the interfaces for the optional parameters for each function :sweat_smile:

jdalrymple:
Feel free to submit a PR if you find the time!! :sunglasses:

PR #606, opened 2020-02-15
feats: adds options interface for MergeRequest.all()

AlvaroBernalG:
#574

### DIFF
--- a/packages/gitbeaker-core/src/services/MergeRequests.ts
+++ b/packages/gitbeaker-core/src/services/MergeRequests.ts
@@ -48,6 +48,30 @@ export interface UpdateMergeRequestOptions {
   allow_maintainer_to_push?: boolean;
 }
 
+export interface AllMergeRequestsOptions {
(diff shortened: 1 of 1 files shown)


jdalrymple:
I know its a pain, but could you also add the type information for the camelCased response. Similar to here: https://github.com/jdalrymple/gitbeaker/blob/gitbeaker-release/packages/gitbeaker-core/src/services/Groups.ts

AlvaroBernalG:
> I know its a pain, but could you also add the type information for the camelCased response. Similar to here: https://github.com/jdalrymple/gitbeaker/blob/gitbeaker-release/packages/gitbeaker-core/src/services/Groups.ts

Sure,

Thanks to @mariusAlch with his https://github.com/MariusAlch/json-to-ts it's going to be a bit less painful 😄

jdalrymple:
:rocket: PR was released in [`19.1.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/19.1.0) :rocket:

PR #639, opened 2020-03-24
Add example code into README.md

akira345:
Hi.
The sample code in the Readme does not appear to have any examples that include inherited members.
Added example using hidden option ``` includeInherited ```.

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,14 +401,16 @@ const api = new Gitlab({
 // Listing users
 let users = await api.Users.all();
 
-// Listing project members(including inherited members)
-// see. https://docs.gitlab.com/ee/api/members.html#list-all-members-of-a-group-or-project-including-inherited-members
-const projectId = 9999;
-let members = await api.ProjectMembers.all(projectId, { includeInherited: true } );
-
 // Or using Promise-Then notation
-api.Projects.all().then((projects) => {
+api.Projects.all().then(projects => {
   console.log(projects);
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -401,6 +401,11 @@ const api = new Gitlab({
 // Listing users
 let users = await api.Users.all();
 
+// Listing project members(including inherited members)
+// see. https://docs.gitlab.com/ee/api/members.html#list-all-members-of-a-group-or-project-including-inherited-members
+const projectId = 9999;
```
Jetersen: does not have to be a projectId could just as well be the group/project ;)

akira345: Sorry , I don't understand well ... 

Should I add sample code that uses ```GroupMembers```? or elase magic number ```projectId```?

ex.
```
// Or using Promise-Then notation
api.Projects.all().then((projects) => {
  console.log(projects);
  // add sample codes.
});
```

Jetersen: For the example I do not think project id is helpful.
Yes you can use project id if you know them by heard or doing a sequence of nested api calls like first getting projects and than using those project to.

My suggestion for the example is to the namespace path so `group/project` or `gitlab-org/gitlab` just like in the url to the project.
https://gitlab.com/gitlab-org/gitlab

akira345: Thank you for pointing this out.
Updated sample code.

PR #646, opened 2020-03-24
Fixed variable name for registry method, as per …

schindld:
…https://docs.gitlab.com/12.8/ee/api/container_registry.html#delete-registry-repository-tags-in-bulk

### DIFF
--- a/packages/gitbeaker-core/src/services/ContainerRegistry.ts
+++ b/packages/gitbeaker-core/src/services/ContainerRegistry.ts
@@ -39,8 +39,7 @@ export class ContainerRegistry extends BaseService {
     const [pId, rId] = [projectId, repositoryId].map(encodeURIComponent);
 
     return RequestHelper.del(this, `projects/${pId}/registry/repositories/${rId}/tags`, {
-      // eslint-disable-next-line @typescript-eslint/camelcase
-      name_regex,
+      nameRegex,
       ...options,
     });
   }
--- a/packages/gitbeaker-core/src/services/ContainerRegistry.ts
+++ b/packages/gitbeaker-core/src/services/ContainerRegistry.ts
@@ -32,13 +32,15 @@ export class ContainerRegistry extends BaseService {
   removeTags(
     projectId: string | number,
     repositoryId: number,
-    tagNameRegex: string,
-    options?: Sudo & { keepN: string; olderThan: string },
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    name_regex: string,
+    options?: Sudo & { keep_n: string; older_than: string },
   ) {
     const [pId, rId] = [projectId, repositoryId].map(encodeURIComponent);
 
     return RequestHelper.del(this, `projects/${pId}/registry/repositories/${rId}/tags`, {
-      tagNameRegex,
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      name_regex,
       ...options,
     });
   }
--- a/packages/gitbeaker-core/src/services/ContainerRegistry.ts
+++ b/packages/gitbeaker-core/src/services/ContainerRegistry.ts
@@ -32,9 +32,8 @@ export class ContainerRegistry extends BaseService {
   removeTags(
     projectId: string | number,
     repositoryId: number,
-    // eslint-disable-next-line @typescript-eslint/camelcase
-    name_regex: string,
-    options?: Sudo & { keep_n: string; older_than: string },
+    nameRegex: string,
+    options?: Sudo & { keepN: string; olderThan: string },
   ) {
     const [pId, rId] = [projectId, repositoryId].map(encodeURIComponent);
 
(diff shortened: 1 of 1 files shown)


Jetersen:
@schindld all variables are meant to be camelCased, we use a method to convert them to snake_casing.

packages/gitbeaker-core/src/services/ContainerRegistry.ts
```diff
@@ -32,13 +32,15 @@ export class ContainerRegistry extends BaseService {
...
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    name_regex: string,
+    options?: Sudo & { keep_n: string; older_than: string },
   ) {
     const [pId, rId] = [projectId, repositoryId].map(encodeURIComponent);
 
     return RequestHelper.del(this, `projects/${pId}/registry/repositories/${rId}/tags`, {
-      tagNameRegex,
+      // eslint-disable-next-line @typescript-eslint/camelcase
+      name_regex,
```
Jetersen: Please remove the disable of lint
```suggestion
      nameRegex,
```

is another way to solve it:
```suggestion
      nameRegex: tagNameRegex,
```

packages/gitbeaker-core/src/services/ContainerRegistry.ts
```diff
@@ -32,13 +32,15 @@ export class ContainerRegistry extends BaseService {
   removeTags(
     projectId: string | number,
     repositoryId: number,
-    tagNameRegex: string,
-    options?: Sudo & { keepN: string; olderThan: string },
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    name_regex: string,
+    options?: Sudo & { keep_n: string; older_than: string },
```
Jetersen:
```suggestion
    nameRegex: string,
    options?: Sudo & { keepN: string; olderThan: string },
```

jdalrymple:
Its done that way because javascript should** be camelCased

packages/gitbeaker-core/src/services/ContainerRegistry.ts
jdalrymple: The first would be cleaner, just changing the name of the variable

Jetersen:
@schindld you do not have the close a pull request, you can use force push or just add more commits to the branch.

schindld:
I suppose the variable name change (`tagNameRegex` to `nameRegex`) is still valid. I'll re-open. 

By the way, gitlab 12.9 introduced a few changes to the container registry api: https://gitlab.com/gitlab-org/gitlab/-/commit/ce99813cf545d5ce4644583a824e40cf1e622844. For deleting tags in bulk, they added `name_regex_delete` and `name_regex_keep`, and deprecated `name_regex`.

packages/gitbeaker-core/src/services/ContainerRegistry.ts
```diff
@@ -32,13 +32,13 @@ export class ContainerRegistry extends BaseService {
   removeTags(
     projectId: string | number,
     repositoryId: number,
-    tagNameRegex: string,
+    nameRegex: string,
     options?: Sudo & { keepN: string; olderThan: string },
```
schindld: This should work until gitlab removes the deprecated `name_regex`. Just ignore the existence of `name_regex_delete` for now.
```suggestion
    options?: Sudo & { nameRegexKeep: string, keepN: string; olderThan: string },
```

jdalrymple:
I tend to go with updating the api library to match what ever their docs have it at, not so much the version of gitlab (since it would be a pain to maintain each gl version). I realllllly try to avoid versioning in the code outside of the library tags. I just make sure to include a breaking change commit message when adding the feature.

For this case, I would drop nameRegex all together and use the latest nameRegexDelete as mentioned in the docs as a positional parameter. If you want to create a version of the lib that handles nameRegex as well, I would probably do two PR's, which would result in two releases of the library. That way people who need support for the older version of gl, can just use that release tag of the library.

schindld:
I understand you don't want to maintain two versions. I'd suggest you keep `nameRegex` at least until they release gitlab 13.0 (when they'll probably remove `name_regex`). But if you remove `nameRegex` now, you won't have a version of gitbeaker that works with any pre-13 version of gitlab (at least for removing registry tags), because `tagNameRegex` never worked...unless I'm missing somewhere in my debugging that it gets converted.

jdalrymple:
Thats what i meant by two PR's. First one with the nameRegex, second one with the newest version :)

jdalrymple:
first one could be a fix, second a breaking change

schindld:
It's been a while since I used GitHub. Can you squash commits in a PR? If not, I'll create a new PR.

jdalrymple:
Yea no stress!

schindld:
I'm using 
```json
  "dependencies": {
    "@gitbeaker/node": "16.0.0"
  }
```
but am not seeing the change (I'm still seeing `tagNameRegex` in the dist). And this is what I see in package-lock.json:
```json
"@gitbeaker/core": "^15.1.0",
```
Shouldn't the core be 16.0.0 too?

jdalrymple:
Hmm, need to fix that in the release script!

schindld:
Thanks, I see it's fixed in 17.0.1 🎉

PR #647, opened 2020-03-24
feat: Updated an API parameter for GitLab 12.9

schindld:
fix(core): 🐛 Updated variable name for registry `removeTags` method, as per https://docs.gitlab.com/12.9/ee/api/container_registry.html#delete-registry-repository-tags-in-bulk

BREAKING CHANGE: 🧨 A `removeTags` method parameter name changed from nameRegex to nameRegexDelete. Works with GitLab 12.9+.

Added nameRegexKeep parameter.


jdalrymple:
Going to put this on hold until i fix the versioning problem

Issue #710, opened 2020-04-22
 Expected the `options.agent` properties to be `http`, `https` or `http2`, got `_events`


justin-calleja:
**Description**

**Steps to reproduce**

`npm i -g @gitbeaker/cli`

```
# export GITLAB_TOKEN and GITLAB_HOST env vars
$ gitbeaker projects all
```

**Expected behaviour**

The cli should show some meaningful output related to gitlab projects.

**Actual behaviour**

I get this error (note I have `gb-token` and `gb-host` set as env vars and cli outputs them when I use `-g`):

```
gitbeaker projects all
⠋ Calling GitlabTypeError: Expected the `options.agent` properties to be `http`, `https` or `http2`, got `_events`
    at normalizeArguments (/Users/juscal/.nvm/versions/node/v12.14.1/lib/node_modules/@gitbeaker/cli/node_modules/got/dist/source/core/index.js:502:27)
    at normalizeArguments (/Users/juscal/.nvm/versions/node/v12.14.1/lib/node_modules/@gitbeaker/cli/node_modules/got/dist/source/as-promise/core.js:36:31)
    at got (/Users/juscal/.nvm/versions/node/v12.14.1/lib/node_modules/@gitbeaker/cli/node_modules/got/dist/source/create.js:87:39)
    at /Users/juscal/.nvm/versions/node/v12.14.1/lib/node_modules/@gitbeaker/cli/node_modules/@gitbeaker/node/dist/index.js:1:2178
    at /Users/juscal/.nvm/versions/node/v12.14.1/lib/node_modules/@gitbeaker/cli/node_modules/@gitbeaker/node/dist/index.js:1:1669
    at Object.next (/Users/juscal/.nvm/versions/node/v12.14.1/lib/node_modules/@gitbeaker/cli/node_modules/@gitbeaker/node/dist/index.js:1:1774)
    at /Users/juscal/.nvm/versions/node/v12.14.1/lib/node_modules/@gitbeaker/cli/node_modules/@gitbeaker/node/dist/index.js:1:711
    at new Promise (<anonymous>)
    at i (/Users/juscal/.nvm/versions/node/v12.14.1/lib/node_modules/@gitbeaker/cli/node_modules/@gitbeaker/node/dist/index.js:1:456)
    at /Users/juscal/.nvm/versions/node/v12.14.1/lib/node_modules/@gitbeaker/cli/node_modules/@gitbeaker/node/dist/index.js:1:2049
```

[...]

PR #709, opened 2020-04-22
Reach > 90% coverage and add Integration Testing

jdalrymple:
- Added many many more unit tests to increase coverage to 90% +
- Removed the upsert functionality of PushRules to maintain documented API parity

fixes #686 
fixes #698
fixes #309

PR #711, opened 2020-04-22
Adding CLI Integration Tests

jdalrymple:
Sister PR to #709 while investigating  #710

### DIFF
--- a/packages/gitbeaker-cli/test/integration/index.ts
+++ b/packages/gitbeaker-cli/test/integration/index.ts
@@ -20,7 +20,7 @@ function cli(cmd, options = {}) {
 }
 
 beforeEach(() => {
-  env = {};
+  env = process.env;
 });
 
 // describe('gitbeaker -g -- CLI global Enviroment Variables', () => {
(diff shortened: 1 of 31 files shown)


jdalrymple:
Getting a weird [error](https://gitlab.com/jdalrymple/gitbeaker/-/jobs/522895624#L43):

```
/bin/sh: node: not found
```
on all the tests EXCEPT the one testing the version: `gitbeaker -v`

jetersen:
Based on the test name I presume your clearing the environment variables. Hence you accidentally cleared PATH env.

jetersen:
providing an empty env object can be dangerous 🗡️

jdalrymple:
True!!!!!!!!!!!! That would make sense. I wonder why that doesnt happen locally though

packages/gitbeaker-cli/test/integration/index.ts
```diff
@@ -20,7 +20,7 @@ function cli(cmd, options = {}) {
 }
 
 beforeEach(() => {
-  env = {};
+  env = process.env;
```
jetersen, line 23: This would not reset it?

I think you need declare it once outside the anonymous function.

jdalrymple: I did this on purpose, because i modify the env var passed each time, i want to reset it before each test.

jdalrymple:
New random error: https://gitlab.com/jdalrymple/gitbeaker/-/jobs/523978110#L37

jdalrymple:
:rocket: PR was released in [`19.0.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/19.0.0) :rocket:

PR #737, opened 2020-04-29
Add types field to node and browser package.json

danielmoore:
Typings for `@gitbeaker/{node,browser}@>= 15` do not work because the `index.d.ts` is in a non-standard location and the `types` field is not set in `package.json`. Interestingly, the `types` field _is_ set in `@gitbeaker/core`.

This change brings all three packages into alignment with the `types` field in both relative location in `package.json` as well as value.

### DIFF
--- a/packages/gitbeaker-browser/package.json
+++ b/packages/gitbeaker-browser/package.json
@@ -49,6 +49,7 @@
   ],
   "license": "MIT",
   "browser": "dist/index.js",
+  "types": "dist/index.d.ts",
   "repository": {
     "type": "git",
     "url": "https://github.com/jdalrymple/gitbeaker"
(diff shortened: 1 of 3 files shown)


jdalrymple:
:rocket: PR was released in [`19.2.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/19.2.0) :rocket:

uinz:
browser type file should be `node_modules/@gitbeaker/browser/dist/types/index.d.ts`

btw, all borwser type is `any`

--- 

I handled this temporarily
```ts
// typings/gitbeaker-browser.d.ts
declare module "@gitbeaker/browser" {
  export * from "@gitbeaker/core/dist/types/index";
}
```

jdalrymple:
@uinz I see that now! Fixing it momentarily.

PR #761, opened 2020-05-05
Add Prettierignore file

jdalrymple:
The lerna config file re-lints every time there is a publish. Since this happens in the publish pipeline, it causes future builds to crash as it does not match prettier's conventions.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+lerna.json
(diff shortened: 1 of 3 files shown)


jdalrymple:
:rocket: PR was released in [`19.3.1`](https://github.com/jdalrymple/gitbeaker/releases/tag/19.3.1) :rocket:

Issue #800, opened 2020-05-17
Use custom request options (such as search_namespaces) in Projects API


xatavian:
**Description**
The GITLAB API includes a lot of query parameters for Projects such as search, search_namespaces, visibility, ... Currently the `Projects` service and its `Projects.search` method does not allow optional query parameters. 

In particular, `search_namespaces` is interesting because it would allow one to search for `namespace/project` instead of just `project` (eg. `jdalrymple/gitbeaker` instead of just `gitbeaker`). 

**Proposal**

Add `BaseRequestOptions` to `Projects.search`: 

```
class Projects extends BaseService {
  // ...
  search(projectName: string, options?: BaseRequestOptions);
}
```

xatavian:
By the way, that could be nice if `projectName` was URI-encoded in search as well. Otherwise, we get weird request URLs :smile:

jdalrymple:
This is an easy one to fix up. Thanks for pointing it out!

xatavian:
Well, I can probably make the change if you'd like :smile:

xatavian:
I have trouble setting up my project to use a local installation of gitbeaker though. I make `@gitbeaker/node` point to my local installation but then it requires a remote `@gitbeaker/core` which actually is the real package that needs change. Should I change the core's package.json as well ?

jdalrymple:
Hehe yea it can be a little confusing. I should add this to the docs. You just have to update @gitbeaker/core, thats where all the logic is. The steps:

1. Update core
2. run yarn build / npm run build
3. go into the gitbeaker-node folder and run npm link
4. test

PR #801, opened 2020-05-17
Feature: Arbitrary parameters for Projects.search

xatavian:
As discussed in #800, the Projects API's search method was missing optional query parameters. This MR allows one to put other optional parameters such as `search_namespaces`.

In #800, I discussed the possibility of URI-encode the search parameter. Turns out it's already done by the RequestHelper and the `query-string` package :smile: 

So, here is my small contribution to the project.

### DIFF
--- a/packages/gitbeaker-core/src/services/Projects.ts
+++ b/packages/gitbeaker-core/src/services/Projects.ts
@@ -124,8 +124,10 @@ export class Projects extends BaseService {
     return RequestHelper.del(this, `projects/${pId}/fork`, options);
   }
 
-  search(projectName: string) {
-    return RequestHelper.get(this, 'projects', { search: projectName });
+  search(projectName: string, options?: BaseRequestOptions) {
+    return RequestHelper.get(this, 'projects', { search: projectName, ...options }) as Promise<
+      ProjectSchema[]
+    >;
   }
 
   share(


packages/gitbeaker-core/src/services/Projects.ts
```diff
@@ -124,8 +124,10 @@ export class Projects extends BaseService {
     return RequestHelper.del(this, `projects/${pId}/fork`, options);
   }
 
-  search(projectName: string) {
-    return RequestHelper.get(this, 'projects', { search: projectName });
+  search(projectName: string, options?: BaseRequestOptions) {
+    return RequestHelper.get(this, 'projects', { search: projectName, ...options }) as Promise<
+      ProjectSchema[]
```
jdalrymple, line 129: I would add this return information to the function header instead of casting the return value. Besides that LGTM!!

xatavian: Ah I see ! I just did the same as in the all method but you're right :p

xatavian: Actually it is not that simple: 

```
search(projectName: string, options?: BaseRequestOptions): Promise<ProjectSchema[]>{
    return RequestHelper.get(this, 'projects', { search: projectName, ...options });
  }
```

raises the following error: `Type 'Promise<GetResponse>' is not assignable to type 'Promise<ProjectSchema[]>'`

jdalrymple: hmm, true. My bad, looks good then

jdalrymple:
:rocket: PR was released in [`19.7.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/19.7.0) :rocket:

Issue #924, opened 2020-06-26
Create todos for issues


s-kazuki:
**Description**
GitLab API supports creating todos for issues: [docs](https://docs.gitlab.com/ee/api/issues.html#create-a-todo)
but gitbeaker supports only for merge requests.

**Proposal**
add method like:
```
+  createIssue(projectId: string | number, issueId: number, options?: Sudo) {
+    return RequestHelper.post(
+      this,
+      `projects/${projectId}/issuess/${issueId}/todo`,
+      options,
+    );
+  }
```
at Todos service

jdalrymple:
I would modify the implementation, ill explain in the PR

PR #925, opened 2020-06-26
:up: create todos for issues

s-kazuki:
ref https://github.com/jdalrymple/gitbeaker/issues/924

### DIFF
--- a/packages/gitbeaker-core/src/services/Todos.ts
+++ b/packages/gitbeaker-core/src/services/Todos.ts
@@ -9,27 +9,19 @@ export class Todos extends BaseService {
     return RequestHelper.get(this, 'todos', options);
   }
 
-  createMergeRequest(projectId: string | number, mergerequestId: number, options?: Sudo) {
-    return RequestHelper.post(
-      this,
-      `projects/${projectId}/merge_requests/${mergerequestId}/todo`,
-      options,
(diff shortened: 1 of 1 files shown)


jdalrymple:
Not sure this structure is the best. Currently the docs have the functionality explained in each of the specific resources (issue and merge requests) but i feel like moving the createTodo function there could be misleading when you know there is a Todo API.

Having the separate functions, although clear, can be a bit misleading which the term "create[Issue/MergeRequest]" since its not creating either of those, but rather a Todo. 

Another idea would maybe be modifying current create function

```
create(projectId, resourceId, { resourceName: 'mergerequest', ...options })
```

Though its a bit more complex, it maintains the understanding that a todo will be created, ensures that the resourceId parameter is still required, and keeps the same default functionality that is creation on mergerequests. Internally we could change the url it points to depending on the resourceName parameter?

Thoughts?

s-kazuki:
@jdalrymple sounds good.
how about my additional commit??

jdalrymple:
LGTM, ill merge it up this weekend

jdalrymple:
:rocket: PR was released in [`23.0.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/23.0.0) :rocket:

PR #1057, opened 2020-08-09
fix(node): response body needs parse firstly when it's type is json

Gkxie:
If the `Content-Type` is `application/json` in response headers, we can't get the error message without parse it.
https://github.com/jdalrymple/gitbeaker/blob/949171bd23bbcfc815c8d8af3ec0fd9a7be5f6eb/packages/gitbeaker-node/src/GotRequester.ts#L62


Gkxie:
@jdalrymple

jdalrymple:
:rocket: PR was released in [`23.6.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/23.6.0) :rocket:

PR #1112, opened 2020-08-24
FormData import not working

jdalrymple:
## Summary
The export of the FormData library seems to cause a problem when in combination of the latest rollup [release](https://github.com/rollup/rollup/releases/tag/v2.26.4).

- import FromData from 'form-data' doesnt work "form_data_1.default is not a constructor"
- import * as FormData from 'form-data' doesnt work since the bundled output isnt of type constructor (its of type object)
- import FormData = require('form-data') doesnt work since require isnt supported in es modules
- const FormData = require('form-data') doesnt work because require isnt supported in the browser.

I havent received any applicable feedback form-data/form-data/issues/470, form-data/form-data/issues/441

## Possible solutions
- this [may](https://github.com/form-data/form-data/pull/471) help if it ever gets merged. Not sure how to go about fixing this :/
- downgrade rollup to before 2.26, though this will only work short term
- :nauseated_face: fork form-data

PR #1113, opened 2020-08-24
feat: apply exponental backoff for 429 errros

### LINKED ISSUE #960: [429 Too Many Requests] How to solve that?
Reference to #73, I still get lots of `429 Too Many Requests` using `api.ProjectImportExport.schedule(id)` and `api.ProjectImportExport.download(id)`.

The error is catch using `catch()` but there is no way to see the `Retry-After`.

From #73, it seems that should be handled inside Gitbeaker. Am I wrong?

Thanks.

max-wittig:
This was implemented similarly to https://github.com/python-gitlab/python-gitlab/blob/1606310a880f8a8a2a370db27511b57732caf178/gitlab/__init__.py#L549
Gitlab sadly doesn't send out `Retry-After` headers anymore, so we need this workaround

Fixes #960

### DIFF
--- a/packages/gitbeaker-browser/src/KyRequester.ts
+++ b/packages/gitbeaker-browser/src/KyRequester.ts
@@ -5,6 +5,7 @@ import {
   DefaultRequestOptions,
   createInstance,
   defaultRequest as baseDefaultRequest,
+  wait,
 } from '@gitbeaker/requester-utils';
 
 function responseHeadersAsObject(response): Record<string, string> {
@@ -53,18 +54,26 @@ export function processBody(response) {
 }
 
 export async function handler(endpoint, options) {
+  const obeyRateLimit = true;
+  const maxRetries = 10;
   let response;
 
-  try {
-    response = await ky(endpoint, options);
-  } catch (e) {
-    if (e.response) {
-      const output = await e.response.json();
-
-      e.description = output.error || output.message;
+  for (let i = 0; i < maxRetries; i += 1) {
+    try {
+      response = await ky(endpoint, options); // eslint-disable-line
+      break;
+    } catch (e) {
+      const waitTime = 2 ** i * 0.1;
+      if (e.response) {
+        if (obeyRateLimit && e.response.status === 429) {
+          await wait(waitTime); // eslint-disable-line
+          continue; // eslint-disable-line
+        }
+        const output = await e.response.json(); // eslint-disable-line
+        e.description = output.error || output.message;
+      }
+      throw e;
     }
-
-    throw e;
   }
 
   const { status } = response;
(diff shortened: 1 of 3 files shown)


max-wittig:
@jdalrymple This might be implemented kind of dirty, but I'm not an expert in JS. If you could suggest some improvements, that would be nice!

jdalrymple:
Awesome!! Ill read it over in the next couple days :smiley:

packages/gitbeaker-node/src/GotRequester.ts
```diff
@@ -51,18 +52,29 @@ export function processBody({ rawBody, headers }: { rawBody: Buffer; headers: ob
 }
 
 export async function handler(endpoint, options) {
```
jdalrymple, line 54: I think for got, we just need to increase the retry limit, it handled the other logic internally https://github.com/sindresorhus/got/blob/master/source/core/calculate-retry-delay.ts

packages/gitbeaker-browser/src/KyRequester.ts
```diff
@@ -53,18 +54,26 @@ export function processBody(response) {
 }
 
 export async function handler(endpoint, options) {
+  const obeyRateLimit = true;
```
jdalrymple, line 57: Similar with ky https://github.com/sindresorhus/ky/#retry

packages/gitbeaker-node/src/GotRequester.ts
jdalrymple: https://github.com/sindresorhus/got#retry

packages/gitbeaker-browser/src/KyRequester.ts
max-wittig: I think that doesn't work, because GitLab doesn't actually send any headers with `Retry-Later`. That's why it doesn't work

packages/gitbeaker-node/src/GotRequester.ts
max-wittig: I think that doesn't work, because GitLab doesn't actually send any headers with `Retry-Later`. That's why it doesn't work. Got already has 10 retries by default.

max-wittig:
@jdalrymple I've been using #1113 and #1116 together since a while in production. Works just fine. Wanna take another look?

jdalrymple:
Ill merge them ASAP. Im currently waiting on #1112 to be merged

jdalrymple:
Finally got the build to pass, going to merge this in the next couple days after resolving the conflicts

max-wittig:
Thanks 👍

jdalrymple:
haha not completely merged yet, there was a hiccup with the build. Sorting it out now though!

PR #1116, opened 2020-08-25
fix(core): ignore header links

### LINKED ISSUE #1115: Keyset pagination is not supported
Keyset pagination is required, when you are requesting more than `50.000` projects.

When using keyset pagination with `gitbeaker`, you'll only get the first page back: 

```js
gitlab.Projects.all({'per_page': 100, 'statistics': true, 'pagination': 'keyset', 'order_by': 'id'})
```

https://docs.gitlab.com/ee/api/#pagination

**Expected behaviour**

gitbeaker should return more than 100 projects, when using keyset pagination

**Actual behaviour**

gitbeaker returns only 100 projects, when using keyset pagination

**Possible fixes**

Not sure. Something with the headers changed at GitLab and `gitbeaker` most likely is reading the wrong headers

max-wittig:
Fixes #1115

### DIFF
--- a/packages/gitbeaker-core/src/infrastructure/RequestHelper.ts
+++ b/packages/gitbeaker-core/src/infrastructure/RequestHelper.ts
@@ -68,14 +68,12 @@ async function get(
     totalPages: parseInt(headers['x-total-pages'], 10),
   };
 
-  const underLimit = maxPages ? pagination.current < maxPages : true;
-
   // Camelize response body if specified
   if (service.camelize) body = camelizeKeys(body);
 
-  // Rescurse through pagination results
-  if (!query.page && underLimit && pagination.next) {
-    const { next } = parseLink(headers.link);
+  // Recourse through pagination results
+  const { next } = parseLink(headers.link);
+  if (!query.page && next) {
     const leaf = service.url.split('/').pop() || '';
     const regex = new RegExp(`.+/api/v\\d(/${leaf})?/`);
     const more = (await get(service, next.replace(regex, ''), {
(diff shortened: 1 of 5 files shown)


packages/gitbeaker-core/src/infrastructure/RequestHelper.ts
```diff
@@ -68,14 +68,12 @@ async function get(
...
 
-  const underLimit = maxPages ? pagination.current < maxPages : true;
-
   // Camelize response body if specified
   if (service.camelize) body = camelizeKeys(body);
 
-  // Rescurse through pagination results
-  if (!query.page && underLimit && pagination.next) {
-    const { next } = parseLink(headers.link);
+  // Recourse through pagination results
```
jdalrymple, line 74: Haha, oops, its suppose to be recurse

packages/gitbeaker-core/src/infrastructure/RequestHelper.ts
```diff
@@ -68,14 +68,12 @@ async function get(
     totalPages: parseInt(headers['x-total-pages'], 10),
   };
 
-  const underLimit = maxPages ? pagination.current < maxPages : true;
```
jdalrymple, line 71: woudldnt this completely ignore the maxPages limit?

max-wittig: That's true, but there is no way to know which the current page is from there. We could add some counter somewhere maybe? Not sure where to put that though.

jdalrymple:
Closing in favour of #1184 , I tried to rebase your branch and it wouldn't let me push, im guessing due to restriction of force pushes which makes sense. I added more tests and updated the docs as well

PR #1186, opened 2020-10-06
Adding pagination tests

### LINKED ISSUE #1176: Result set is limited to 20 objects since version 23.5
**Description**
It seems, that since version 23.5, the `gitlab.[...].all(...)` methods are only receiving 20 result object per default. Since version 23.5, I have to add the `perPage` option, to get more results.

(Tested in a pure browser environment)

**Steps to reproduce**
* Call `gitlab.Issues.all()` (on a project with more then 20 Issues
* Length of resulting array is exactly 20.

**Expected behaviour**
* Per default, all available objects should be part of the result.

jdalrymple:
fixes #1176


jdalrymple:
:rocket: PR was released in [`24.0.1`](https://github.com/jdalrymple/gitbeaker/releases/tag/24.0.1) :rocket:

PR #1327, opened 2020-11-12
fix: call `buildMap` dynamically instead of using json file

kiprasmel:
questionable change.

I suppose the static json file is used to create it once to avoid extra startup cost?

I changed this because I'm trying my hardest to setup webpack to compile both my project and your's as a git submodule, and I somehow made it work, but I do need this work-around for it. I am fine as I have my fork, but ideally I could base on top of the original repo.

Any thoughts?

### DIFF
--- a/packages/gitbeaker-cli/src/cli.ts
+++ b/packages/gitbeaker-cli/src/cli.ts
@@ -2,7 +2,7 @@ import * as Sywac from 'sywac';
 import * as Chalk from 'chalk';
 import { camelize, decamelize, depascalize } from 'xcase';
 import * as Gitbeaker from '@gitbeaker/node';
-import * as map from '@gitbeaker/core/dist/map.json';
+import { buildMap } from "@gitbeaker/core/../scripts/generate"
 
 // Styling settings
 const commandStyle = Chalk.hex('#e34329').bold;
@@ -200,7 +200,7 @@ cli.command('*', (argv, ctx) => {
 });
 
 // Add all supported API's
-Object.entries(map).forEach(([apiName, methods]) => {
+Object.entries(buildMap()).forEach(([apiName, methods]) => {
   cli.command(param(apiName), {
     desc: `The ${apiName} API`,
     setup: (setupArgs) => setupAPIs(setupArgs, apiName, methods),
--- a/packages/gitbeaker-core/scripts/generate.ts
+++ b/packages/gitbeaker-core/scripts/generate.ts
@@ -35,7 +35,7 @@ function removeOptionalArg(list) {
   return list;
 }
 
-function buildMap() {
+export function buildMap() {
   const map = {};
   const baseArgs = Object.keys(getParamNames(BaseService)[0]);
 


jdalrymple:
Its done this way for exactly the reason you mention. It would be silly for it to be auto generated at runtime since nothing about the file changes after the initial compilation. In the past i looked into finding a way to not have to export it in such an odd manner, but i couldnt find anything :(

kiprasmel:
Actually now that I think about it, you should be able to inline it at build time with your module bundler (so that the current `buildMap()` would be replaced with the json that it returns) -- you can do this with [webpack's "define plugin"](https://webpack.js.org/plugins/define-plugin/), but don't know about rollup since I haven't used it much.

webpack example:


```js
// webpack.config.js

const webpack = require("webpack");
const { buildMap } = require(...);

module.exports = {
	// ...,
	plugins: [
	new webpack.DefinePlugin({
		// https://webpack.js.org/plugins/define-plugin/#runtime-values-via-runtimevalue
		__builtMap: webpack.DefinePlugin.runtimeValue(
			() => buildMap(),
			true // webpack will rebuild if any file change occurs
		),
	}),
}

```

and you'd use it like so (`gitbeaker-cli/src/cli.ts`):

```diff
-import * as map from '@gitbeaker/core/dist/map.json';

// Add all supported API's
-Object.entries(map).forEach(([apiName, methods]) => {
+Object.entries(__builtMap).forEach(([apiName, methods]) => {
  cli.command(param(apiName), {
    desc: `The ${apiName} API`,
    setup: (setupArgs) => setupAPIs(setupArgs, apiName, methods),
  });
});

export { cli };
```

while also making it type-safe:

```ts
// gitbeaker-cli/index.d.ts

declare const __builtMap: string;
```

and it'd work! (I have the same setup with just different variables in https://github.com/kiprasmel/refined-gitlab. Don't know about rollup though:D

jdalrymple:
Could maybe make use of this [plugin](https://github.com/aMarCruz/rollup-plugin-jscc)

kiprasmel:
I think this is the one: https://github.com/rollup/plugins/tree/master/packages/replace

jdalrymple:
> Actually now that I think about it, you should be able to inline it at build time with your module bundler (so that the current `buildMap()` would be replaced with the json that it returns) -- you can do this with [webpack's "define plugin"](https://webpack.js.org/plugins/define-plugin/), but don't know about rollup since I haven't used it much.
> 
> webpack example:
> 
> ```js
> // webpack.config.js
> 
> const webpack = require("webpack");
> const { buildMap } = require(...);
> 
> module.exports = {
> 	// ...,
> 	plugins: [
> 	new webpack.DefinePlugin({
> 		// https://webpack.js.org/plugins/define-plugin/#runtime-values-via-runtimevalue
> 		__builtMap: webpack.DefinePlugin.runtimeValue(
> 			() => buildMap(),
> 			true // webpack will rebuild if any file change occurs
> 		),
> 	}),
> }
> ```
> 
> and you'd use it like so (`gitbeaker-cli/src/cli.ts`):
> 
> ```diff
> -import * as map from '@gitbeaker/core/dist/map.json';
> 
> // Add all supported API's
> -Object.entries(map).forEach(([apiName, methods]) => {
> +Object.entries(__builtMap).forEach(([apiName, methods]) => {
>   cli.command(param(apiName), {
>     desc: `The ${apiName} API`,
>     setup: (setupArgs) => setupAPIs(setupArgs, apiName, methods),
>   });
> });
> 
> export { cli };
> ```
> 
> while also making it type-safe:
> 
> ```ts
> // gitbeaker-cli/index.d.ts
> 
> declare const __builtMap: string;
> ```
> 
> and it'd work! (I have the same setup with just different variables in https://github.com/kiprasmel/refined-gitlab. Don't know about rollup though - any chance for you to try and find something similar?

Here how are you importing the build function? Since its a ts file?

jdalrymple:
So i managed to pull it off but its rough. Right now it accomplishes it by:

1. transpile generate code from ts to esm using ts-node
2. runs generate code to create map file using esm 
3. rules overall rollup with replace that does a require of this map file and embeds it
4. deleted the generated map file

This is quite inefficent but i couldnt think of a quicker way of just running the buildMap function within the replace logic in the rollup config *because* the buildMap function imports the whole API for parsing which is ts and rollup does not understand ts without transpiling.

A better way of accomplishing this would be using AST and getting the exported classes/instance functions and their property names. 

If you can think of any enhancements of this method, id love to hear them!

PR #1352, opened 2020-11-27
Export the APIMap through a compile-time replacement

jdalrymple:
A new function will be exported from the core called getAPIMap, this will allow for the import of api mapping without needing to look at the dist folder.

See #1327 for discussion

To ensure this didnt mess with the typing, can someone give it a run and make sure the typing for the exported services appear correctly?

<details>
  <summary>📦 Published PR as canary version: <code>25.6.1-canary.1352.225920336.0</code></summary>
  <br />
  
  :sparkles: Test out this PR locally via:
  
  ```bash
  npm install @gitbeaker/browser@25.6.1-canary.1352.225920336.0
  npm install @gitbeaker/cli@25.6.1-canary.1352.225920336.0
  npm install @gitbeaker/core@25.6.1-canary.1352.225920336.0
  npm install @gitbeaker/node@25.6.1-canary.1352.225920336.0
  npm install @gitbeaker/requester-utils@25.6.1-canary.1352.225920336.0
  # or 
  yarn add @gitbeaker/browser@25.6.1-canary.1352.225920336.0
  yarn add @gitbeaker/cli@25.6.1-canary.1352.225920336.0
  yarn add @gitbeaker/core@25.6.1-canary.1352.225920336.0
  yarn add @gitbeaker/node@25.6.1-canary.1352.225920336.0
  yarn add @gitbeaker/requester-utils@25.6.1-canary.1352.225920336.0
  ```
</details>

### DIFF
--- a/packages/gitbeaker-cli/src/cli.ts
+++ b/packages/gitbeaker-cli/src/cli.ts
@@ -2,7 +2,7 @@ import * as Sywac from 'sywac';
 import * as Chalk from 'chalk';
 import { camelize, decamelize, depascalize } from 'xcase';
 import * as Gitbeaker from '@gitbeaker/node';
-import * as map from '@gitbeaker/core/dist/map.json';
+import { getAPIMap } from '@gitbeaker/core';
 
 // Styling settings
 const commandStyle = Chalk.hex('#e34329').bold;
@@ -200,7 +200,7 @@ cli.command('*', (argv, ctx) => {
 });
 
 // Add all supported API's
-Object.entries(map).forEach(([apiName, methods]) => {
+Object.entries(getAPIMap()).forEach(([apiName, methods]) => {
   cli.command(param(apiName), {
     desc: `The ${apiName} API`,
     setup: (setupArgs) => setupAPIs(setupArgs, apiName, methods),
--- a/packages/gitbeaker-core/rollup.config.js
+++ b/packages/gitbeaker-core/rollup.config.js
@@ -1,3 +1,4 @@
+import replace from '@rollup/plugin-replace';
 import pkg from './package.json';
 import { commonConfig, commonPlugins } from '../../rollup.config';
 
@@ -14,5 +15,10 @@ export default {
       format: 'es',
     },
   ],
-  plugins: commonPlugins,
+  plugins: [
+    replace({
+      __apiMap__: JSON.stringify(require('./dist/map.json'))
+    }),
+    ...commonPlugins,
+  ]
 };
--- a/packages/gitbeaker-core/src/index.ts
+++ b/packages/gitbeaker-core/src/index.ts
@@ -3,7 +3,7 @@ import * as APIServices from './services';
 
 /* -------------- Export Map ------------- */
 export function getAPIMap() {
-  return JSON.parse('__apiMap__')
+  return JSON.parse('__apiMap__');
 }
 
 /* -------------- Single Services ------------- */
--- a/packages/gitbeaker-core/src/index.ts
+++ b/packages/gitbeaker-core/src/index.ts
@@ -1,6 +1,11 @@
 import { bundler } from './infrastructure';
 import * as APIServices from './services';
 
+/* -------------- Export Map ------------- */
+export function getAPIMap() {
+  return JSON.parse('__apiMap__')
+}
+
 /* -------------- Single Services ------------- */
 export * from './services';
 
--- a/packages/gitbeaker-core/src/index.ts
+++ b/packages/gitbeaker-core/src/index.ts
@@ -2,7 +2,11 @@ import { bundler } from './infrastructure';
 import * as APIServices from './services';
 
 /* -------------- Export Map ------------- */
-export function getAPIMap() {
+export function getAPIMap(): Record<string, unknown> {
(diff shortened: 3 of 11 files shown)


jdalrymple:
@kiprasmel Would this work for your use case as well?

kiprasmel:
Great work! Yea I think this should work, will try it out soon. I think there's still an improvement we can make - hold on a sec

packages/gitbeaker-core/src/index.ts
```diff
@@ -1,6 +1,15 @@
 import { bundler } from './infrastructure';
 import * as APIServices from './services';
 
+/* -------------- Export Map ------------- */
+export function getAPIMap(): Record<string, unknown> {
+  if (!'__apiMap__'.includes('{')) {
+    throw new Error('This function is only available in the distributed code');
+  }
+
+  return JSON.parse('__apiMap__');
```
kiprasmel, line 10: shouldn't `__apiMap__` be a global variable instead of a string? That's how it works with webpack, but not sure about rollup.

packages/gitbeaker-core/rollup.config.js
```diff
@@ -14,5 +15,10 @@ export default {
       format: 'es',
     },
   ],
-  plugins: commonPlugins,
+  plugins: [
+    replace({
+      __apiMap__: JSON.stringify(require('./dist/map.json'))
```
kiprasmel, line 20: Couldn't we also parse the JSON right here to avoid doing it in runtime? It would be faster afaik

packages/gitbeaker-cli/src/cli.ts
```diff
@@ -200,7 +200,7 @@ cli.command('*', (argv, ctx) => {
 });
 
 // Add all supported API's
-Object.entries(map).forEach(([apiName, methods]) => {
+Object.entries(getAPIMap()).forEach(([apiName, methods]) => {
```
kiprasmel, line 203: As opposed to using the `getAPIMap` function, what if we just had here the `__apiMap__` constant that contains already valid & parsed json as a javascript object to achieve the best performance?

```suggestion
Object.entries(__apiMap__).forEach(([apiName, methods]) => {
```

kiprasmel:
About the issue that the script to generate the `map.json` is in typescript -- I don't actually know how you work around it here, but my idea would be to create a wrapper script that would call `ts-node` on the script that generates the map, taking the output and putting it all in the `__apiMap__` variable at compile time.

```js
plugins: [
    replace({
      __apiMap__: JSON.parse(JSON.stringify(require('./wrapper-script.js').generateAPIMapWithTSNode())),
	}),
```

packages/gitbeaker-core/src/index.ts
jdalrymple: The rollup version does a string replace at compile time

packages/gitbeaker-core/rollup.config.js
jdalrymple: It does a string replace, not a variable replace. So if i tried to return a object, it would be an object wrapped in quotes. Unless im missing something

packages/gitbeaker-cli/src/cli.ts
jdalrymple: See my previous points

jdalrymple:
> About the issue that the script to generate the `map.json` is in typescript -- I don't actually know how you work around it here, but my idea would be to create a wrapper script that would call `ts-node` on the script that generates the map, taking the output and putting it all in the `__apiMap__` variable at compile time.
> 
> ```js
> // rollup.config.js
> 
> plugins: [
>     replace({
>       __apiMap__: require('./wrapper-script.js').generateAPIMapWithTSNode(),
> 	}),
> ```
> 
> ```js
> // wrapper-script.js
> const { spawnSync } = require('child_process');
> 
> export function generateAPIMapWithTSNode() {
> 	spawnSync('yarn generate-map');
> 	const generatedMap = JSON.parse(require('./dist/map.json'));
> 	return generatedMap;
> }
> ```
> 
> and then the global `__apiMap__` variable would contain an object of the generated & parsed map.

This is what is done in the npm script. See generate script

kiprasmel:
Oh, now I understand. Thank you for your patience:D

Apart from this type error, I think everything should be fine:

```console
[tsl] ERROR in /home/kipras/projects/refined-gitlab/gitbeaker/packages/gitbeaker-core/scripts/generate.ts(15,3)
      TS2322: Type 'false | string[]' is not assignable to type 'string[]'.
  Type 'false' is not assignable to type 'string[]'.

```

https://github.com/jdalrymple/gitbeaker/blob/8b197bac4fded5d8aeb17dc366018b0a65e05a87/packages/gitbeaker-core/scripts/generate.ts#L14-L23

jdalrymple:
Should be good now @kiprasmel , give it a peak when you get a moment

kiprasmel:
@jdalrymple rebased my fork on top of this PR, compiled and tried out refined-gitlab -- works just fine! 🎉

jdalrymple:
:rocket: PR was released in [`26.0.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/26.0.0) :rocket:

kiprasmel:
Awesome!

PR #1411, opened 2020-12-27
Fixing scoping problem changing Service url argument to prefixUrl

### LINKED ISSUE #1399: Internal class extender is using a global reference to a closure variable
Receiving 404 when using GroupMembers.all(groupId);
The same group id works for Groups.show(groupId);
When performing a manual GET request to https://gitlab.com/api/v4/groups/groupid/members/all I'm getting a valid response.
```
const gitlab = Gitlab({token: "TOKEN"})
gitlab.GroupMembers.all(groupId); // Throws an error.

await axios.get("https://gitlab.com/api/v4/groups/10450319/members", { headers: { "private-token": ACCESS_TOKEN },}); // Works fine
```

I've noticed that it tries to request the api without the `/groups/` section.

jdalrymple:
fixes #1399

### DIFF
--- a/packages/gitbeaker-core/src/services/Commits.ts
+++ b/packages/gitbeaker-core/src/services/Commits.ts
@@ -55,8 +55,8 @@ export interface CommitAction {
 }
 
 export interface GPGSignature {
-  signature_type: "PGP",
-  verification_status: "verified"|"unverified";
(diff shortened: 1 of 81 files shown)


jdalrymple:
I split this into two PRs

PR #1467, opened 2021-01-19
Remove possibility of char type information polluting check of content-type

### LINKED ISSUE #1418: Endpoints return strings instead of objects
**Steps to reproduce**

```ts
import * as gitlab from '@gitbeaker/node';

const api = new gitlab.Gitlab({
    host: 'https://gitlab.example.com',
    token: '123456789123456789123456789',
});

void (async () => {
    const group = await api.Groups.show(1);

    console.log(typeof group);
})();
```

```
experiments on  master [✘!?] via ⬢ v14.15.3
➜ ts-node gitlab.ts
string
```

**Expected behaviour**

That it returns me an object with of the type `GroupDetailSchema`.

**Actual behaviour**

That it returns me a `string`

jdalrymple:
fixes #1418

### DIFF
--- a/packages/gitbeaker-browser/src/KyRequester.ts
+++ b/packages/gitbeaker-browser/src/KyRequester.ts
@@ -42,7 +42,8 @@ export function defaultOptionsHandler(
 }
 
 export function processBody(response) {
-  const contentType = response.headers.get('content-type') || '';
+  // Split to remove potential charset info from the content type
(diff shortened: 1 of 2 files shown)


jdalrymple:
:rocket: PR was released in [`28.0.3`](https://github.com/jdalrymple/gitbeaker/releases/tag/28.0.3) :rocket:

PR #1531, opened 2021-02-05
chore: add `root: true` to eslint

kiprasmel:
I still use eslint v5 because I want support for my `~/.eslintrc` (and because I cannot be bothered to update lol), so I couldn't commit because eslint would complain until I did this.

See also https://eslint.org/docs/user-guide/migrating-to-7.0.0#runtime-deprecation-warnings-for-eslintrc-config-files


jdalrymple:
:rocket: PR was released in [`28.1.1`](https://github.com/jdalrymple/gitbeaker/releases/tag/28.1.1) :rocket:

PR #1604, opened 2021-03-14
Experiment with Cloud run to spin up test gitlab instances

jdalrymple:
Updates the pipeline to utilize an external container for the gitlab instance


jdalrymple:
There seems to be a problem with the gitlab service used in the CI script. I'm not sure how to sort this out. Locally it seems to work just fine and without access to the service logs from the CI doesnt aid in finding the problem

jdalrymple:
I need to find a better solution for this. Integration tests running on gitlab ci seem to still be temperamental. Testing them locally everything seems fine :/

PR #1633, opened 2021-03-19
Fix #1624 - Fixed mime type handling

### LINKED ISSUE #1624: Problem fetching image files
Hi, I've encountered a problem when trying to download image files from a repository using the `nodejs` version of gitbeaker and the `RepositoryFiles.showRaw` function. I've been able to track the problem down to this switch statement 

https://github.com/jdalrymple/gitbeaker/blob/4b1ea4afc400c7913559f95624d63cd9c32a99ed/packages/gitbeaker-node/src/GotRequester.ts#L54-L66

Basically the result (e.g. `image/png`) falls into the default code block and is converted to string. But to me, the switch statement seems wrong - it is treating any non-matching mime as a string, which would cause problems with many other filetypes. Treating only the text mime type and its subtypes as string and defaulting to buffer for everything else seems like a much better idea. I know I can just write my own requester function (and I'll probably do that in the meantime), but I feel like this should be fixed upstream. I will be glad to make a PR :slightly_smiling_face:

Pryx:
Fixes the problem reported in issue #1624 and adds appropriate unit tests.

### DIFF
--- a/packages/gitbeaker-browser/src/KyRequester.ts
+++ b/packages/gitbeaker-browser/src/KyRequester.ts
@@ -45,19 +45,15 @@ export function processBody(response) {
   // Split to remove potential charset info from the content type
   const contentType = (response.headers.get('content-type') || '').split(';')[0].trim();
 
-  switch (contentType) {
-    case 'application/json': {
-      return response.json().then((v) => v || {});
-    }
-    case 'application/octet-stream':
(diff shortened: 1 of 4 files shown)


jdalrymple:
:rocket: PR was released in [`28.2.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/28.2.0) :rocket:

PR #1790, opened 2021-05-21
refactor: remove camelization from type system

zmmille2:
Removes camelization from being represented in the type system
Updates BaseService to remove generic parameter
Updates types leveraging BaseService to remove generic camelize paramter

In general, it seems like the API can return different casings in different responses.
If that's the case, handling it when processing requests and responses seems to me to
be a better approach that letting it propagate throughout the type system.

This may be a breaking change, as if I am understanding it correctly, objects before
this pr had two properties: `fooBar` and `foo_bar`.

Hopefully, after this change, this will not be the case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,9 +41,11 @@
     "lint:doc": "prettier --check './*.json' './*.yml' './*.js'",
     "lint:doc:fix": "prettier --write './*.json' './*.yml' './*.md' './*.js'",
     "lint:fix": "prettier --write 'packages/**/{src,test}/**/*.ts' && eslint 'packages/**/{src,test}/**/*/*.ts' --fix",
+    "test:int": "yarn test:integration",
     "test:integration": "jest test/integration",
     "test:integration:node": "TEST_ID=$(uuid) yarn jest node/test/integration",
     "test:unit": "jest test/unit",
+    "test": "yarn test:unit && yarn:integration",
     "release": "auto shipit --verbose --verbose"
   },
   "author": "Justin Dalrymple <justin.s.dalrymple@gmail.com>",
--- a/packages/gitbeaker-core/src/infrastructure/RequestHelper.ts
+++ b/packages/gitbeaker-core/src/infrastructure/RequestHelper.ts
@@ -1,7 +1,7 @@
 import { parse as parseLink } from 'li';
 import { camelizeKeys } from 'xcase';
 import { BaseService } from '@gitbeaker/requester-utils';
-import { appendFormFromObject, Camelize } from './Utils';
+import { appendFormFromObject } from './Utils';
 
 // Request Options
 export type IsForm = {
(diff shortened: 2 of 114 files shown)


jdalrymple:
I may be missing some of your logic here. The response objects only have one or the other, not both. With this PR, you completely remove the correct typing that would occur if you did pass in a camelize configuration. Correct me if I'm wrong though?

zmmille2:
I don't think I am - `baseService` still takes `camelize` as an argument:
https://github.com/jdalrymple/gitbeaker/blob/bc57c9191fc7a28d9ec4888570dac983820b214b/packages/gitbeaker-requester-utils/src/BaseService.ts#L32-L46

It just isn't a _type_ argument anymore.

I'm proposing that you parse the response and represent it as the same object internally, rather than duplicating properties.
I might be missing the correct configuration for where `camelize` gets passed in, but it should be updated to be passed as a part of `options`.

zmmille2:
I know this is a massive PR. The very substantive changes are in `package.json` and `RequestHelper.ts`. I believe everything else is similar. Hope that helps the review.

jdalrymple:
I think this is where the misunderstanding it:

>This may be a breaking change, as if I am understanding it correctly, objects before
this pr had two properties: fooBar and foo_bar.

^ That shouldnt be happening, nor does it happen with imports excluding the bundles (this is a bug)

When you pass true for that camelize property, the responses you receive are typed to be camelized. You can see this by looking at any of the services (excluding the bundles: Gitlab, ProjectsBundle, UsersBundle, GroupsBundle). For example:

camelized
not-camelized

zmmille2:
This is the bit I'm keyed in on:
> When you pass true for that camelize property, the responses you receive are typed to be camelized. You can see this by looking at any of the services (excluding the bundles: Gitlab, ProjectsBundle, UsersBundle, GroupsBundle). For example:
> 
> camelized
> not-camelized

This seems strange to me. Even if the responses you send out are camelized, why support both in your internal models? Why not just process objects as you're sending or receiving them? I think it would simplify typing, which is why I've made this PR.

jdalrymple:
Hmm, could you show me an example? I'm not sure what you mean by supporting both internally ? The only support that is added internally is the typing. Hit me with an example of what you mean haha

zmmille2:
`RequesteHelper.ts` already has this snippet:
(in `master`):
https://github.com/jdalrymple/gitbeaker/blob/master/packages/gitbeaker-core/src/infrastructure/RequestHelper.ts#L101

(in this branch):
https://github.com/jdalrymple/gitbeaker/blob/bc57c9191fc7a28d9ec4888570dac983820b214b/packages/gitbeaker-core/src/infrastructure/RequestHelper.ts#L90

seems like the response can be camelized or not, which this PR still supports.

> The only support that is added internally is the typing.
Yes, this is the bit that I think you should drop. Why support snake_case and camelCase? Is there a specific use case you're targeting? Why not support expand support to UPPERCASE or lowercase as properties, for example?

I'm not sure what convention you might be following, but I think it is fairly standard practice to use lowerCamelCase. [Google recommends it, for example](https://google.github.io/styleguide/jsguide.html#naming-non-constant-field-names).

Which, now that I stop and think about things, means that I really think that all instances of `snake_case` should be replaced with `camelCase` in internal models... which would certainly be a hefty refactor.

Am I making sense here?

jdalrymple:
I think I'm following. 

No where in the code, internally, am i "supporting" snake_case. 

Yes that code snippet does control whether the consumer receives a camelCased response or the default snake_case response from the Gitlab API. The only additional "support" is reflecting this within the types. So while you did not remove support for the camelizing of the response, you removed support for the type representation of this configuration. 

Why did I use snake_cased as the default types? This is because they represent EXACTLY what is returned from Gitlab making it much easier to debug and meets the default expectation from users who want to use the Gitlab API. For example when you are reading to see what options are available through the gitlab docs, the names will match by default, unless you add that camelize option. Its also easier to go from snake_case to camelCamel than camelCase to snake_case (example would be for names like myVARname -> my_v_a_r_name vs my_var_name -> myVarName)

Yes I know I could stick to the JS standard and just export everything in camelCase, but I've had instances in the past where consumers wanted the exact output from Gitlab which was the original support of this library. The camelCase support was added later. so, as a happy balance and to avoid breaking their experience, support for both is present. 

Further, while it may not be present right now, the eventual goal is to pull the API schema structures directly from Gitlab once they expose their OpenAPI spec, which would require the same support. :sunglasses:  planning for the future here haha

zmmille2:
Sounds intentional and great!

PR #1791, opened 2021-05-21
feat: export models directly through index for use in consumer code

### LINKED ISSUE #1218: Export Schema type
**Description**
You should expore type of schema,  so that others  can import. 

**Proposal**
For example,  I have  to  do this:

``` js 
import { GroupSchema } from "@gitbeaker/core/dist/types/services/Groups";
  public async fetchSubGroups(options: Sample): Promise<GroupSchema> {
    const { id } = options;
    const result = await this.core.Groups.all({ id });
    return result;
  }
```

zmmille2:
Move Schemae out of their respective files and into a models subdirectory
Reconcile duplicates
Add model index to export models
Add model export to core to provide access to consumers.

Fixes #1218.


jdalrymple:
This is good! I'm currently testing something similar right now which offers a little more control over the exports. The decision to have the types located where they are directly used was intentional, though I did originally have many of the types in this format. I appreciate you taking the time to move this issue forward!!

jdalrymple:
On second thought, lemme think over it for a bit.

zmmille2:
No worries! Let me know if there’s any adjustments I could make that might make sense. Just wanted to try to help out 👍

jdalrymple:
So i may do something similar to this, the exports will be in a different format though:

```ts
import {
  API_MAP,
  API_TYPES,
  
    Gitlab,
  ...all the bundles,

  Projects,
  ...all the single APIs
} from '@gitbeaker/core'

const {ProjectSchema} = API_TYPES // Not married to this naming convention yet haha, still thinking about this way of implementation. I dont want to make the change and then have it cause me a headache in the future lol
```

because gitbeaker/node, gitbeaker/cli and gitbeaker/browser use the imported services  and having those ALSO mixed with the types would be alot of unnecessary parsing.

jdalrymple:
I could explicitly import each of the resources in the other libraries :s, it would just suck.

zmmille2:
From what I'm reading here, it sounds like this change's general thrust will be accepted in another state. Feel free to close, build on this, whatever, and feel free to ping me if there are adjustments I can make here.

jdalrymple:
Closing in favor of #1818

PR #1889, opened 2021-06-21
Improve pagination logic by parsing query parameters explicitly on response

### LINKED ISSUE #1880: Pagination is broken
**Description**

I'm trying to do a .all query to get all results but it seems to be just sending the request for the same page to the gitlab server and I'm waiting indefinately on my issues.

When I added a maxPages=3, I noticed that I do get 3 times my requested page size back, but also 3x the same issues from the first page.

**Steps to reproduce**
```ts
  const api = new Gitlab({token/host});
  const project = await api.Projects.show("projectpath");
  let issues = await api.Issues.all({ projectId: project.id, status: 'opened', perPage: 10});
```

**Expected behaviour**

Get all pages with the relevant information

**Actual behaviour**

Get the same page over and over again (when maxPages is set, if not it just waits till the end of time).

This is all with the latest. I had to revert to release 28.4.1 couldn't get any newer to work. Unfortunately now I don't have typescript type support anymore :-(

jdalrymple:
fixes #1880

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -12,6 +12,7 @@
     "@gitbeaker/requester-utils": "^30.2.0",
     "form-data": "^4.0.0",
     "li": "^1.3.0",
+    "query-string": "^7.0.0",
     "xcase": "^2.0.1"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


jdalrymple:
:rocket: PR was released in [`30.3.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/30.3.0) :rocket:

PR #1932, opened 2021-07-10
Added Topics to ProjectSchema

Aliyss:


Aliyss:
Also maybe something to consider, currently ``ProjectExtendedSchema`` has the ``namespace``. However when testing ``namespace`` already returned without authorization. Thus it should be in the ``ProjectSchema`` right?

Aliyss:
I'll just keep adding stuff as I use the repo... Feel free to merge whenever you feel like.

jdalrymple:
> `ProjectExtendedSchema`

yup! That should be changed to match the return data for sure

packages/core/src/resources/RepositoryFiles.ts
```diff
@@ -109,12 +109,13 @@ export class RepositoryFiles<C extends boolean = false> extends BaseResource<C>
...
     const [pId, path] = [projectId, filePath].map(encodeURIComponent);
 
-    return RequestHelper.get()(this, `projects/${pId}/repository/files/${path}/raw`, {
-      ref,
-      ...options,
-    }) as unknown as Promise<Blob>;
+    return RequestHelper.get()(
+      this,
+      `projects/${pId}/repository/files/${path}/raw`,
+      options,
```
Aliyss, line 118: @jdalrymple Could you check if this is ok? I am quite sure that ref isn't mandatory, but I don't know if options still needs the spread syntax.

jdalrymple: Youre right! https://docs.gitlab.com/ee/api/repository_files.html#get-raw-file-from-repository

jdalrymple:
:rocket: PR was released in [`32.0.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/32.0.0) :rocket:

PR #1937, opened 2021-07-13
Add interface schemas to MergeRequestApprovals

### LINKED ISSUE #1936: MergeRequestApprovals.approvalRules returns incorrect type
**Description**

When using `MergeRequestApprovals.approvalRules`, TypeScript shows the type as `Record<string, unknown>` but the function actually returns an array of approval rules (`Record<string, unknown>[]`). I'm going to submit a PR to fix this but wanted to file an issue first.

**Steps to reproduce**

Use `MergeRequestApprovals.approvalRules` function with a project id.

**Expected behaviour**

Return an array of Approval Rules. This is expected, however TypeScript thinks it is just a `Record`, so it throws an error when trying to access array members.

**Actual behaviour**

TS throws an error.

**Possible fixes**

Add a ApprovalRules interface. This is probably the best solution. I'll look into it.

The bandaid fix is just to define the method to return an array of `Record<string, unknown>` like `Record<string, unknown>[]`.

mattwilkinsonn:
Adds Schemas to the methods on MergeRequestApprovals.

 I also renamed some functions:
- `editApprovals` -> `editApprovalsRequired` 
- `approvers` to `addApprovers` 
- `approvals` to `approvalConfig`
to make their usage more clear.

`approvalConfig` I also changed slightly to not be able to query by a merge request ID. /approvals on a project vs on a merge request return very differently. I added the `approvalStatus` method to query the /approvals endpoint on a particular merge id, you can see the docs on this [here](https://docs.gitlab.com/ee/api/merge_request_approvals.html#get-configuration-1).

Fixes #1936


packages/core/src/resources/MergeRequestApprovals.ts
```diff
@@ -57,25 +95,27 @@ export class MergeRequestApprovals<C extends boolean = false> extends BaseResour
...
     }
-    return RequestHelper.get()(this, url, options);
+    return RequestHelper.get<MergeRequestApprovalRuleSchema[]>()(this, url, options);
   }
 
-  approvals(
-    projectId: string | number,
-    { mergerequestIid, ...options }: { mergerequestIid?: number } & BaseRequestOptions = {},
-  ) {
+  approvalConfig(projectId: string | number, options?: BaseRequestOptions) {
```
jdalrymple, line 101: This is a good change, def improved readability. The only caveat is that it may be misleading for those looking for the approvalConfig for merge requests. If you looked at the docs you'd assume a similar API, but its a different function entirely. I think we could still do something similar to what was done before, but with a inferred response based on the input argument.  I know i do it somewhere else in the codebase, but id have to double check and follow up!

packages/core/src/resources/MergeRequestApprovals.ts
```diff
@@ -143,25 +187,27 @@ export class MergeRequestApprovals<C extends boolean = false> extends BaseResour
...
-    return RequestHelper.put()(this, url, { name, approvalsRequired, ...options });
+    return RequestHelper.put<MergeRequestApprovalRuleSchema>()(this, url, {
+      name,
+      approvalsRequired,
+      ...options,
+    });
   }
 
-  editApprovals(
+  editApprovalsRequired(
```
jdalrymple, line 197: Id change this to editApprovalConfig for consistency

jdalrymple: Also, why remove the support for the project level edits? https://docs.gitlab.com/ee/api/merge_request_approvals.html#change-configuration

jdalrymple:
Thanks for the updates! I read it over and added some comments. Tell me what you think.

jdalrymple:
Gonna jump on this one this week :crossed_fingers:

mattwilkinsonn:
Sorry, been super busy with work stuff and moving apartments the last few weeks. Reviewing your feedback now.

packages/core/src/resources/MergeRequestApprovals.ts
mattwilkinsonn: Added `editProjectApprovalConfig` for the project level edits. The name may need to be changed but I don't think the MR and Project approval configs should be in the same method as they take very different parameters. Let me know what you think.

packages/core/src/resources/MergeRequestApprovals.ts
mattwilkinsonn: Ok interesting, if you point me to that i'll take a look

jdalrymple:
No hurry! I jumped on this one pretty hard, im just fixing up some of the type discrimination and ill push the updates.

jdalrymple:
Here is a sample:

```ts
  configuration(
    projectId: string | number,
    options: BaseRequestOptions,
  ): Promise<ProjectLevelMergeRequestApprovalSchema>;

  configuration(
    projectId: string | number,
    options: { mergerequestIid: number } & BaseRequestOptions,
  ): Promise<MergeRequestLevelMergeRequestApprovalSchema>;

  configuration(
    projectId: string | number,
    { mergerequestIid, ...options }: { mergerequestIid?: number } & BaseRequestOptions = {},
  ) {
    let url: string

    if (mergerequestIid) {
      const mIid = encodeURIComponent(mergerequestIid);
      url =  `projects/${pId}/merge_requests/${mIid}/approvals`
    } else {
      url = `projects/${pId}/approvals`
    }

    return RequestHelper.get()(this, url, options);
  }
```

Im using overload methods to handle the typing properly

jdalrymple:
Hmm, or maybe i should just split it into two different API imports, though that seems like a bit of overkill

PR #2025, opened 2021-08-23
fix(core): change Commits.diff response type to CommitDiffSchema[]

F3n67u:
see https://docs.gitlab.com/ee/api/commits.html#get-the-diff-of-a-commit


jdalrymple:
:rocket: PR was released in [`32.2.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/32.2.0) :rocket:

PR #2056, opened 2021-09-03
Update Jobs showPipelineJobs to have correct array return type

lemkepf:
Getting the Pipeline Jobs is an array of jobs, not a single job. This is a breaking change. Better than having to do something like: 

``` typescript
var manualJobsResponse = await gitlabApi.Jobs.showPipelineJobs(projectId, pipeline.id, {
    scope: 'manual'
})

var manualJobs = manualJobsResponse as unknown as JobSchema[];
```


jdalrymple:
:rocket: PR was released in [`34.0.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/34.0.0) :rocket:

PR #2070, opened 2021-09-10
Users.search should return an array

sagacity:
The GitLab users/search endpoint returns an array, not a single user.


jdalrymple:
:rocket: PR was released in [`34.2.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/34.2.0) :rocket:

Issue #2069, opened 2021-09-10
protect branch implementation wrong


WangLeto:
https://github.com/jdalrymple/gitbeaker/blob/1f0f4514f241626c73834256525d1e63c8eede16/packages/core/src/resources/ProtectedBranches.ts#L39

Currently the params is sent as POST request body, but refer to Gitlab [API doc](https://docs.gitlab.com/ee/api/protected_branches.html#protect-repository-branches), the params should be sent as url query string.

PR #2142, opened 2021-10-12
Fix a query formatting issue

nilennoct:
There is a mistake when formatting query. For example:

```ts
// Throws "refs is invalid"
gitlab.Repositories.mergeBase(this.repository, ['a', 'b'])
```

GitLab accepts query like `refs[]=a&refs[]=b`, but currently it is formatted to `refs[0]=a&refs[1]=b`.

PR #2155, opened 2021-10-17
Fixing protect request format

jdalrymple:
fixes: #2069

### DIFF
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -15,7 +15,7 @@
     "ky": "0.28.6"
   },
   "devDependencies": {
-    "@rollup/plugin-commonjs": "^20.0.0",
+    "@rollup/plugin-commonjs": "^21.0.0",
     "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-node-resolve": "^13.0.0",
     "@rollup/plugin-replace": "^3.0.0",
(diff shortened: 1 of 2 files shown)


jdalrymple:
:rocket: PR was released in [`34.4.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/34.4.0) :rocket:

jdalrymple:
Query just adds those properties as a query param since normally post requests only send a body.

Looking at this example

```
curl --request POST --header "PRIVATE-TOKEN: <your_access_token>" "https://gitlab.example.com/api/v4/projects/5/protected_branches?name=*-stable&push_access_level=30&merge_access_level=30&unprotect_access_level=40"
```

The props are sent as query params, so that piece of code should work. Ill double check it though

GMZwinge:
Got it. The problem may not necessarily be due to this specific pull request, but we see the problem below with `34.4.0` and we don't see it with the previous `34.2.0`.

With this set of options:
```
{
  allowed_to_push: [
    {
      access_level: 0,
    },
    {
      user_id: <aPushUserId>,
    },
  ],
  allowed_to_merge: [
    {
      access_level: 0,
    },
    {
      user_id: <aMergeUserId>,
    },
  ],
  allow_force_push: false,
  code_owner_approval_required: false,
}
```
The GitLab UI shows the protect branch with the above user checked, but no role checked when it should have the role "No one" checked.

GMZwinge:
Could https://github.com/jdalrymple/gitbeaker/pull/2142 be the reason?

jdalrymple:
Maybe! Ill test it out

PR #2173, opened 2021-10-23
Refactor endpoint generation, using tagged template to encode parameters

nilennoct:
Use tagged template to simplify endpoint generating. For example:

```ts
// Before
const [pId, path] = [projectId, filePath].map(encodeURIComponent);
const url = `projects/${pId}/repository/files/${path}`;

// After
const url = endpoint`projects/${projectId}/repository/files/${filePath}`;
```

### DIFF
--- a/packages/core/src/infrastructure/Utils.ts
+++ b/packages/core/src/infrastructure/Utils.ts
@@ -31,3 +31,19 @@ export function getAPIMap(): Record<string, unknown> {
     throw new Error('This function is only available in the distributed code');
   }
 }
+
+/**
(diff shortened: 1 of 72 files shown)


jdalrymple:
Cool function! The only additional thing id ask for is a basic unit test for the endpoint function. Just because it will be used everywhere, a double check that its always working will be handy.

nilennoct:
I've rebased the master branch and pushed a commit with basic unit test for the `endpoint` function.

jdalrymple:
:rocket: PR was released in [`34.5.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/34.5.0) :rocket:

Issue #2161, opened 2021-10-18
Add support for `include_merged_yaml` on Lint API


MartinHowarth:
**Description**

https://docs.gitlab.com/ee/api/lint.html#yaml-expansion (specifically the YAML expansion) is not supported in this package currently, as far as I can tell. This feature proposal adds support for it.

**Proposal**

It should be as simple as adding the option to pass `include_merged_yaml: boolean` to the existing code [here](https://github.com/jdalrymple/gitbeaker/blob/master/packages/core/src/resources/Lint.ts#L12) and parsing `merged_yaml` back out of the response.

I'm extremely new to TypeScript so I haven't just submitted a PR as I thought it would be more effort for you to coach me through fixing the PR up than to just add the feature yourself! Please shout if I've got that wrong and I'll look into making an MR.

jdalrymple:
Hey! This would be a perfect small PR. All you have to change is the type parameter from Sudo, to BaseRequestOptions, and update the LintSchema to optionally include the merged_yaml (use a ?)

PR #2185, opened 2021-10-28
Add ability to request merged_yaml from the Gitlab Lint API

MartinHowarth:
Fix for issue https://github.com/jdalrymple/gitbeaker/issues/2161

Haven't figured out how to test this yet, any advice on whether it's worth making a UT for this would be useful. Need to work out how to use a branch package to test this live as well - will get to that in a week or two most likely.

### DIFF
--- a/packages/core/src/resources/Lint.ts
+++ b/packages/core/src/resources/Lint.ts
@@ -1,14 +1,15 @@
 import { BaseResource } from '@gitbeaker/requester-utils';
-import { RequestHelper, Sudo } from '../infrastructure';
+import { BaseRequestOptions, RequestHelper } from '../infrastructure';
 
 export interface LintSchema extends Record<string, unknown> {
   status: string;
   errors?: string[];
   warnings?: string[];
+  merged_yaml?: string;
 }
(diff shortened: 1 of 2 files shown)


jdalrymple:
Looks good to me! 

Hmm to test this, since it depends on what Gitlab actually supports, you could write an integration test which calls this function and check that the returned object includes that new key. An example would be like this https://github.com/jdalrymple/gitbeaker/blob/master/packages/node/test/integration/resources/Projects.ts#L14

Does that make sense?

MartinHowarth:
Thank you for your patience.

Have added an integration test for this - hopefully that just works, otherwise I'll try to work out how to run the tests locally to fix it.

jdalrymple:
Looks good! Ill run it through the pipeline to see how those tests run

jdalrymple:
:rocket: PR was released in [`34.5.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/34.5.0) :rocket:

MartinHowarth:
Thanks so much for finishing this!

PR #2222, opened 2021-11-19
Add support for CI lint with namespace context

MartinHowarth:
Add support for CI lint with namespace context. i.e. this section of the API docs: https://docs.gitlab.com/ee/api/lint.html#validate-a-ci-yaml-configuration-with-a-namespace

As before, I don't know how to test this manually, so hopefully the integration test I've added works!


packages/core/src/resources/Lint.ts
```diff
@@ -7,9 +7,28 @@ export interface LintSchema extends Record<string, unknown> {
...
 
 export class Lint<C extends boolean = false> extends BaseResource<C> {
   lint(content: string, options?: BaseRequestOptions) {
+    // Perform CI file linting without context.
+    // See https://docs.gitlab.com/ee/api/lint.html#validate-the-ci-yaml-configuration
+    // This API doesn't work for CI files that contain `local` includes. Use `lint_with_namespace` instead.
     return RequestHelper.post<LintSchema>()(this, 'ci/lint', { content, ...options });
   }
+
+  lint_with_namespace(projectId: string | number, content: string, options?: BaseRequestOptions) {
```
jdalrymple, line 25: I would actually just embed this in the previous function with a check for the context. Ill give it a quick modification and push up!

jdalrymple:
Looks good otherwise!

packages/core/src/resources/Lint.ts
MartinHowarth: Good idea! Slight problem because the two different endpoints return slighty different data. Namely "status" vs "valid". We could coerce the results to look the same by taking the "status" from the without-namespace response and change it to "valid: true|false" depending on it's value? That choice really depends on your philosophy of how accurate this repo is intended to be compared to the gitlab API.

Without namespace:
```
{
  "status": "valid",
  "errors": [],
  "warnings": []
  "merged_yaml": "---\n:test_job:\n  :script: echo 1\n"
}
```

With namespace:
```
{
  "valid": true,
  "merged_yaml": "---\n:test_job:\n  :script: echo 1\n",
  "errors": [],
  "warnings": []
}
```

jdalrymple:
```ts
export interface LintSchema extends Record<string, unknown> {
  status: string;
  errors?: string[];
  warnings?: string[];
  merged_yaml?: string;
}

export interface ContextualLintSchema extends Omit<LintSchema, 'status'> {
  valid: boolean;
}
```
I realize for now, with how ts is, i cant actually maintain the type separation and combine the functions. so keep them separate for now i suppose. just switch the function name to be camelCase.

MartinHowarth:
Thank you! 😄

jdalrymple:
:rocket: PR was released in [`34.7.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/34.7.0) :rocket:

PR #2296, opened 2021-12-23
Fix camelize types

toniopelo:
Currently resources types are not camelized as they should be when the lib is instantiated with the `camelize` option.
```ts
  import { Gitlab } from '@gitbeaker/browser'

  const gl = new Gitlab({
    token,
    host: 'http://smthg.example',
    camelize: true,
  })

  // The return type here is not camelized
  return gl.Users.current()
```

This is because the type Camelize is "lost" in the chain and does not reach the resource types, this PR fixes this typing problem.

### DIFF
--- a/packages/core/src/resources/Gitlab.ts
+++ b/packages/core/src/resources/Gitlab.ts
@@ -302,7 +302,9 @@ const resources = {
   Wikis,
 };
 
-export class Gitlab<C extends boolean = false> extends (class {} as new () => BundledService) {
+export class Gitlab<C extends boolean = false> extends (class {} as new <
+  Camelize extends boolean = false,
+>() => BundledService<Camelize>)<C> {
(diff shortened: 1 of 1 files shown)


toniopelo:
Hi @jdalrymple, this one is types related only and contains really small changes. I took the liberty to create it without prior approval because this is only a fix of what seemed to be the intended behavior. Don't hesitate to tell me if I'm wrong :)

jdalrymple:
Is the problem only seen in the 'Gitlab' resource? or in all them :s

jdalrymple:
Ill include this in #2258

jdalrymple:
:rocket: Issue was released in [`37.0.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/37.0.0) :rocket:

PR #2326, opened 2022-01-17
Fix type of the customAttributeId

4ops:
Set `string` type for the custom attribute key, as described in [docs](https://docs.gitlab.com/ee/api/custom_attributes.html#single-custom-attribute).


jdalrymple:
Good catch! This will be included in #2258

PR #2424, opened 2022-03-19
Update del requests to accept body parameters

### LINKED ISSUE #2319: RepositoryFiles.remove throws an error on success
**Description**

Hi there,
It looks like there is a bug in parsing the response when running the `RepositoryFiles.remove` action.
When running the method, an error is thrown despite the fact that the method did succeed deleting a file from GitLab.

reference to the GitLab-related docs - [here](https://docs.gitlab.com/ee/api/repository_files.html#delete-existing-file-in-repository).
*Worth mentioning that when trying the native REST API from Postman, the response is 204
image

*note that also for the library does succeed in deleting the file despite the error.

**Steps to reproduce**

in your code, use the following lines:
```
     import { Gitlab } from '@gitbeaker/browser';
     const api = new Gitlab({ token: '<your-token>', host: '<your-host>' });
     return await api.RepositoryFiles.remove(projectId, filePath, branch, message);
```

**Expected behaviour**

The promise will resolve properly if the deletion was successful and the file was actually deleted. 

**Actual behaviour**

Instead of resolving, an error is thrown on an non-parsed response:
```
SyntaxError: Unexpected end of JSON input
    at eval (index.js?bb11:31)
    at Generator.next (<anonymous>)
    at eval (index.js?bb11:30)
    at new Promise (<anonymous>)
    at __awaiter (index.js?bb11:30)
    at processBody (index.js?bb11:31)
    at eval (index.js?bb11:31)
    at Generator.next (<anonymous>)
    at fulfilled (index.js?bb11:30)
```
the error makes it difficult to understand if the file was deleted without actually issuing another call to check if there is a file.

**Possible fixes**

Perhaps when the response is 204 from GitLab, the library should just resolve the promise.
* If it helps, I can try to issue a PR fix, but I'll need some clues on where to look :)

jdalrymple:
fixes #2319

### DIFF
--- a/packages/core/src/infrastructure/RequestHelper.ts
+++ b/packages/core/src/infrastructure/RequestHelper.ts
@@ -229,9 +229,12 @@ export function del<T extends Record<string, unknown> | void = void>() {
   return async <C extends boolean, E extends boolean = false>(
     service: BaseResource<C>,
     endpoint: string,
-    { sudo, showExpanded, ...query }: BaseRequestOptions & ShowExpanded<E> = {},
+    { sudo, showExpanded, query, ...options }: BaseRequestOptions & ShowExpanded<E> = {},
   ): Promise<ExtendedRecordReturn<C, E, T>> => {
+    const body = options;
+
     const r = await service.requester.delete(endpoint, {
+      body,
       query,
       sudo,
     });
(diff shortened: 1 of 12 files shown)


jdalrymple:
:rocket: PR was released in [`35.6.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/35.6.0) :rocket:

PR #2813, opened 2022-11-03
Expand Groups.search's options parameter type to PaginatedRequestOptions

### LINKED ISSUE #2812: Groups.search options parameter typed only as Sudo but should be PaginatedRequestOptions (or at least BaseRequestOptions)
**Description**

Currently the Groups.search function's options parameter is typed as a Sudo. However, GitLab supports all the same pagination options as the Groups.all function plus extras like "min_access_level". This is understandable because the documentation of https://docs.gitlab.com/ee/api/groups.html#search-for-group is horrible and doesn't let on that you can pass these extra parameters.

As a general comment, I would not type any of the options parameters as just Sudo unless it can be reasoned that query parameters will NEVER make sense for the given GitLab feature (which should be very rare). Typing them them as at least a BaseRequestOptions will allow for users to still use GitLab API query parameters that gitbreaker hasn't caught up with.

**Proposal**

Change the type of the options parameter of the Groups.search function from Sudo to PaginatedRequestOptions:

  search(nameOrPath: string, options?: PaginatedRequestOptions) {
    return RequestHelper.get()(this, 'groups', {
      search: nameOrPath,
      ...options,
    });
  }

**Checklist**
- [X] I have checked that this is not a duplicate issue.
- [X] I have read the documentation.

phillipB-nmprofile:
closes: #2812

### DIFF
--- a/packages/core/src/resources/Groups.ts
+++ b/packages/core/src/resources/Groups.ts
@@ -114,7 +114,7 @@ export class Groups<C extends boolean = false> extends BaseResource<C> {
     );
   }
 
-  search(nameOrPath: string, options?: Sudo) {
+  search(nameOrPath: string, options?: PaginatedRequestOptions) {
     return RequestHelper.get()(this, 'groups', {
       search: nameOrPath,
       ...options,


jdalrymple:
:rocket: PR was released in [`35.8.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/35.8.0) :rocket:

PR #2853, opened 2022-11-25
Typing: make sure CommitSchema type is okay

matchaxnb:
It was set as "Date" except they do not get hydrated on get so they should be string.

In a TypeScript environment, they are taken as date but actually are strings. This commits makes it so they are considered as what they are: strings. Alternative would be to auto-hydrate them through Date.parse but that would be breaking API.


jdalrymple:
I went back and fixed each of these problems throughout the repo in #2258

PR #2866, opened 2022-11-28
Add options to query Deployments more fully. Fix a bug in MergeRequests.

matchaxnb:
fix(merge-requests): fix snake_case naming of order_by prop in MergeRequestsOptions
    feat(deployments): add options for querying deployments

Basically, we want to query deployments by updated date, we now can. And the merge requests options "orderBy" was not gonna work as GitLab consumes "order_by" instead.

### DIFF
--- a/packages/core/src/resources/Deployments.ts
+++ b/packages/core/src/resources/Deployments.ts
@@ -39,8 +39,17 @@ export type DeploymentSchema = {
   environment: EnvironmentSchema;
 };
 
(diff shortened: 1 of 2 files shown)


jdalrymple:
Just to be clear, you could of passed these options previously, this simply added more explicit types for that function correct? Also, the camelCasing of the options interface is intentional, the proper snakecasing is handled in the background before sending to gitlab.

Anyways, I included your explicit types in the #2258 PR! Thanks!!

PR #3222, opened 2023-04-26
v37 Release Housekeeping

### LINKED ISSUE #3223: How do I import types at version 37.0.0
**Description**

How is it possible to import types of api functions like `MergeRequestSchema`.

Since Version 37.0.0 it is not possible to import types as in older versions like the following snippet.
`import {MergeRequestSchema} from "@gitbeaker/core/dist/types/resources/MergeRequests";`

I have tried the following imports:
`import {MergeRequestSchema} from "@gitbeaker/core";`
`import {MergeRequestSchema} from "@gitbeaker/rest";`

The import from `@gitbeaker/core` leads to a compile error:
> "@gitbeaker/core"' has no exported member named 'MergeRequestSchema'

I think the right way is to export all types which are used by public api functions.

**Checklist**

jdalrymple:
- Updating packages and setting all-contributors-cli version for @auto dependency
 - Fixing type conflicts
 - Fixing exposed type exports 

fixes #3223


jdalrymple:
:rocket: PR was released in [`37.1.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/37.1.0) :rocket:

PR #3234, opened 2023-05-03
Update Readme to show Axolo as a sponsor :)

cosydney:


jdalrymple:
I ended up handling this in #3235 since I also needed to update the other sponsor at the same time. I've also added your logo to each of the npm package README's as well.

PR #3268, opened 2023-05-20
Handle casing edge cases

### LINKED ISSUE #3256: Projects.upload() always says "The provided content-type 'text/plain' is not supported."
**Description**

- Node.js version: 18.16.0
- Gitbeaker version: 38.3.0
- Gitbeaker release (cli, node, browser, core, requester-utils): @gitbeaker/rest

When uploading a file via `Projects.upload()`, I always see this error:
```
{
  description: `"The provided content-type 'text/plain' is not supported."`,
  response: Response {
    size: 0,
    timeout: 300000,
    [Symbol(Body internals)]: [Object],
    [Symbol(Response internals)]: [Object]
  }
}
```
I am passing a base64 image like this:
```
const img = 'data:image/png;base64,[...]'; // also tried without "data:image/png;base64,"
const uploadResponse = await this.gitlab.Projects.upload(
  GITLAB_PROJECTID,
  {
    content: new Blob([img], { type: 'image/png' }),
    filename: 'xyz.png',
  },
);
```
type is set to `image/png`. So why is the API saying I'm using `text/plain`, which seems to be not supported?

**Checklist**

### LINKED ISSUE #3263: Issue links API broken
I tried to use the Issue links API, but it's currently broken.

In [IssueLinks.ts](https://github.com/jdalrymple/gitbeaker/blob/a31690e51e5add769a2adbc145bd2620a317dcbb/packages/core/src/resources/IssueLinks.ts#L48) one argument is `targetIssueIId`. When it's converted to snake_case, it is converted to `target_issue_i_id` instead of `target_issue_iid`.  

The resulting url looks like this
```
https://gitlab.com/api/v4/projects/:project_id/issues/:issue_iid/links?target_project_id=0&target_issue_i_id=0
```

## Fix suggestion

Changing the argument to `targetIssueIid` would fix this issue.

Alternatively the conversion from camelCase to snake_case could be changed to only insert underscores before Capital letters, if the letter before is not Capital.  But I think this is unnecessary complex.

Btw thanks a lot for this great project and the very active maintenance!

jdalrymple:
fixes #3263

### DIFF
--- a/packages/core/src/resources/IssueLinks.ts
+++ b/packages/core/src/resources/IssueLinks.ts
@@ -70,7 +70,7 @@ export class IssueLinks<C extends boolean = false> extends BaseResource<C> {
       endpoint`projects/${projectId}/issues/${issueIId}/links`,
       {
         targetProjectId,
-        targetIssueIId,
+        targetIssueIid: targetIssueIId,
         ...options,
       },
     );
--- a/packages/core/src/resources/LinkedEpics.ts
+++ b/packages/core/src/resources/LinkedEpics.ts
@@ -53,7 +53,7 @@ export class LinkedEpics<C extends boolean = false> extends BaseResource<C> {
       {
         searchParams: {
           targetGroupId,
-          targetEpicIId,
+          targetEpicIid: targetEpicIId,
         },
         ...options,
       },


jdalrymple:
:rocket: PR was released in [`38.7.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/38.7.0) :rocket:

PR #3275, opened 2023-05-25
Added fix for Group Access tokens requests

Bambii556:
The url used for group access tokens was wrong, it was using 'Projects' instead of 'Groups'


jdalrymple:
:rocket: PR was released in [`38.9.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/38.9.0) :rocket:

PR #3276, opened 2023-05-26
Fix ExternalStatusChecks arguments

ksilligan:
According API docs https://docs.gitlab.com/ee/api/status_checks.html 
externalCheckStatusId should be passed as "external_status_check_id". 

Additionally, I think show() method should be removed, because it does not do anything


jdalrymple:
:rocket: PR was released in [`38.9.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/38.9.0) :rocket:

PR #3286, opened 2023-06-01
Mark some MR nullable types as nullable

ravewill:
I encountered a null reference error today when using this library. The head_pipeline property on an object implementing
ExpandedMergeRequestSchema was null, and Typescript hadn't hinted that it could be null. Unfortunately GitLab's API docs don't tell you what values could be null, so I did a GET request to the REST endpoint of my self-hosted GitLab server (15.11.3) and updated these type definitions based on null JSON values I got back.

Let me know if this PR is in the right format, or if I need to add anything else! :smile:

### DIFF
--- a/packages/core/src/resources/Issues.ts
+++ b/packages/core/src/resources/Issues.ts
@@ -19,8 +19,8 @@ import type { MetricImageSchema } from './AlertManagement';
 export interface TimeStatsSchema extends Record<string, unknown> {
   time_estimate: number;
   total_time_spent: number;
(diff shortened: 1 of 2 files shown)


ravewill:
@jdalrymple thanks for making it so easy to contribute! :heart:

jdalrymple:
:rocket: PR was released in [`38.10.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/38.10.0) :rocket:

PR #3297, opened 2023-06-06
Switching to a more strict "Omit" type to support better mapped typing

### LINKED ISSUE #2134: Types is incorrect after omitting keys
**Description**

After omitting some fields from existing Schema types, the types of remained fields become `unknown`.

**Steps to reproduce**

```typescript
const api = new Gitlab({ token: 'token' });
const branch = await api.Branches.show('projectId', 'master');
const commitDiff = await api.Commits.diff('projectId', branch.commit.id);
// TS2345: Argument of type 'unknown' is not assignable to parameter of type 'string'.
```

**Expected behaviour**

The type of `branch.commit.id` should be `string`.

**Actual behaviour**

The type of `branch.commit.id` is `unknown`.

**Possible fixes**

All Schema types extend from `Record<string, unknown>`. For example:

```typescript
interface Foo extends Record<string, unknown> {
    x: string;
    y: number;
    z: boolean;
}

type K0 = keyof Foo; // string | number
type K1 = Exclude<K0, 'x'>; // string | number
type T = Omit<Foo, 'x'>; // { [x: string]: unknown, [x: number]: unknown }
```

While the other is not extended from `Record` type:

```typescript
interface Bar {
    x: string;
    y: number;
    z: boolean;
}

type K0 = keyof Bar; // 'x' | 'y' | 'z'
type K1 = Exclude<K0, 'x'>; // 'y' | 'z'
type T = Omit<Bar, 'x'>; // { y: number, z: boolean }
```

Because `Omit` is implemented by `Exclude`, so the excluded keys is incorrect if the object type extends from `Record` type.

jdalrymple:
fixes #2134

### DIFF
--- a/packages/core/src/infrastructure/Utils.ts
+++ b/packages/core/src/infrastructure/Utils.ts
@@ -46,6 +46,8 @@ export type EitherOrNone4<T, U, V, W> =
 
 export type AllOrNone<T extends Record<string, any>> = T | Partial<Record<keyof T, never>>;
 
+export type MappedOmit<T, K extends keyof T> = { [P in keyof T as P extends K ? never : P]: T[P] };
+
 export type OptionValueType = undefined | string | boolean | Blob | number | (Blob | string)[];
 
 export function appendFormFromObject(object: Record<string, OptionValueType>): FormData {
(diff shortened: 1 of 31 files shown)


jdalrymple:
:rocket: PR was released in [`38.12.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/38.12.0) :rocket:

PR #3305, opened 2023-06-08
Fix the pipeline trigger endpoint

### LINKED ISSUE #3236: Pipeline triggering feature was deleted
**Description**

At some point in time of this project, the gitlab API could be called to trigger a pipeline on a project (https://github.com/jdalrymple/gitbeaker/commit/5858fc) , but the project structure has changed a lot since then and the trigger feature seems to have been lost in the changes. 
I would expect to find it here : https://github.com/jdalrymple/gitbeaker/blob/main/packages/core/src/resources/Pipelines.ts or here https://github.com/jdalrymple/gitbeaker/blob/main/packages/core/src/resources/PipelineTriggerTokens.ts in the trigger method, but the trigger method seems to do the same as the create method.

**Proposal**

Add the trigger pipeline for a project following this documentation : https://docs.gitlab.com/ee/ci/triggers/

**Checklist**

jdalrymple:
fixes #3236


jdalrymple:
:rocket: PR was released in [`39.0.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/39.0.0) :rocket:

PR #3317, opened 2023-06-26
Support defered token retrieval 

### LINKED ISSUE #3310: Defer token resolution for when making the call
**Description**

I am trying to write a service with the use of your library. Because it's running at all times I can't hardcode the API token when I first start it up, as we might very well decide to regenerate it for security purposes. Thus I basically want to resolve the API token whenever I am about to send a request.
I am basically looking at how I could pass a callback for a tokenProvider rather than the actual token into the Gitlab() constructor. 
There isn't much documentation around this project and kudos for how extendable the code is but that makes it really tough to read 😅 
Because of the token checks in the BaseResource constructor I would probably need to send a dummy value just to fool it and then maybe play around with sending a requesterFn that would resolve the token and override the value in the headers before sending it down. 
I do kind of feel a bit dirty to do it this way though 😂 and I was curios if you would have any ideas on how maybe to do it in a cleaner way or maybe if you think it's sensible to switch/extend the constructor from accepting token string values to providers. Performance wise it shouldn't impact anyone as a provider like () => "<key_value>" is the same as passing the string down.

Let me know what you think 😁 

**Checklist**

jdalrymple:
closes #3310

### DIFF
--- a/packages/core/test/unit/resources/ApplicationAppearance.ts
+++ b/packages/core/test/unit/resources/ApplicationAppearance.ts
@@ -27,7 +27,7 @@ describe('ApplicationAppearance.edit', () => {
   it('should request PUT /application/appearence without arguments', async () => {
     await service.edit();
 
-    expect(RequestHelper.put()).toHaveBeenCalledWith(service, 'application/appearence', undefined);
+    expect(RequestHelper.put()).toHaveBeenCalledWith(service, 'application/appearence', {});
   });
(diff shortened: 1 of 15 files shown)


jdalrymple:
:rocket: PR was released in [`39.2.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/39.2.0) :rocket:

PR #3360, opened 2023-07-28
Updating protected branched types

### LINKED ISSUE #3358: Type mismatch with Gitlab API for /protected-branches
Hey I noticed a type mismatch with the Gitlab API on the protected-branches endpoint.
In the Gitlab API it seems they return all `access_level` , `user_id` , `group_id` which are essentially 3 types of access. Because a single ExtendedProtectedBranchAccessLevel can only be of one of the 3 types only one will be set out of the 3 the other being returned as null. 
Well in your types you don't expect access_level to be null which it could ~~and also you expect user_id and group_id to be missing which seems can't be the case~~ Sorry I just read that they are not present for non premium users so that makes sense.

[Link to Gitlab docs mentioning this](https://docs.gitlab.com/ee/api/protected_branches.html#:~:text=%7B%0A%20%20%20%20%20%20%22id%22%3A%20%201%2C-,%22access_level%22%3A%20null%2C,-%22user_id%22%3A%20null%2C%0A%20%20%20%20%20%20%22group_id).

https://github.com/jdalrymple/gitbeaker/blob/86724f338e9deee46bcc311a85a56b4c44883207/packages/core/src/resources/ProtectedBranches.ts#L13C4-L13C4

jdalrymple:
fixes #3358

### DIFF
--- a/packages/core/src/resources/ProtectedBranches.ts
+++ b/packages/core/src/resources/ProtectedBranches.ts
@@ -10,7 +10,8 @@ import type {
 } from '../infrastructure';
 
 export type ProtectedBranchAccessLevel = 0 | 30 | 40 | 60;
-export interface ExtendedProtectedBranchAccessLevel {
+export interface ExtendedProtectedBranchAccessLevelSchema {
+  id: number;
   access_level: ProtectedBranchAccessLevel;
   access_level_description: string;
   user_id?: number | null;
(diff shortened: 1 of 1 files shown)


jdalrymple:
:rocket: PR was released in [`39.10.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/39.10.0) :rocket:

PR #3361, opened 2023-07-28
Make camelize type recursive

### LINKED ISSUE #3359: When "camilize" is true, does not have a recursive camelize on types
**Description**

- Node.js version: 18.15.0
- Gitbeaker version: 38.12.1
- Gitbeaker release (cli, node, browser, core, requester-utils): `@gitbeaker/core`(used by `@gitbeaker/rest`)
- OS & version: Arch Linux

When `camelize` is `true`, all object keys, including child objects, are in camel-case. However, the types only recognize camel-cased attributes on the root object.

**Steps to reproduce**

1. Create an `@gitbeaker/rest` instance with the parameter `camelize: true`;
2. Call any endpoint with child objects(for example `MergeRequestApprovals.showApprovalState`);
3. Using an IDE with type check enabled, try to access a child object property that needs to be present in camel-case(for example: `rules[].eligibleApprovers`);
4. The value is accessible because the object is camilzed, but the type does not recognize the camel-case property. The types only recognize `rules[].eligible_approvers`), even if `camelize` is enabled.

**Possible fixes**

1. Recognize both types(with `Camelize<>` and without it);
2. Apply recursive `Camelize<>` type when `camelize: true`.

**Checklist**
- [ ] I have checked that this is not a duplicate issue.
- [ ] I have read the documentation.

jdalrymple:
fixes #3359

### DIFF
--- a/packages/core/src/infrastructure/Utils.ts
+++ b/packages/core/src/infrastructure/Utils.ts
@@ -15,11 +15,9 @@ export type CamelizeString<T extends PropertyKey> = T extends string
     : T
   : T;
 
-export type Camelize<T> = { [K in keyof T as CamelizeString<K>]: T[K] };
+export type Camelize<T> = { [K in keyof T as CamelizeString<K>]: Camelize<T[K]> };
 
-export type Never<T> = {
-  [P in keyof T]?: never;
-};
+export type Never<T> = { [P in keyof T]?: never };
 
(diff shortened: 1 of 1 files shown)


jdalrymple:
:rocket: PR was released in [`39.9.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/39.9.0) :rocket:

PR #3373, opened 2023-08-07
Support for endpoint rate limits

### LINKED ISSUE #3330: Add support for rate limiting requests
**Description**

@jdalrymple In a past issue, you asked me if I was having trouble with timeouts. Well, at that time (pointing to GitLab 15), no.

Unfortunately, hitting a different GitLab 16 server is triggering timeouts after between 4 and 7 pages of results.

Do you know of anything I can dig into and/or is there a backoff mechanism somewhere?

**Checklist**

jdalrymple:
closes #3330

### DIFF
--- a/packages/requester-utils/src/BaseResource.ts
+++ b/packages/requester-utils/src/BaseResource.ts
@@ -1,4 +1,4 @@
-import { RequesterType, ResourceOptions, RateLimitOptions } from './RequesterUtils';
+import { RateLimitOptions, RequesterType, ResourceOptions } from './RequesterUtils';
 
 export interface RootResourceOptions<C> {
   // TODO: Not actually optional - Need to fix wrapper typing in requestUtils.ts:
--- a/packages/requester-utils/src/BaseResource.ts
+++ b/packages/requester-utils/src/BaseResource.ts
@@ -40,7 +40,7 @@ function getDynamicToken(tokenArgument: (() => Promise<string>) | string): Promi
 // Default rate limits per minute
 const DEFAULT_RATE_LIMITS = Object.freeze({
   // Default rate limit
-  '*': 30,
+  '*': 3000,
 
   // Import/Export
   'projects/import': 6,
--- a/packages/requester-utils/src/BaseResource.ts
+++ b/packages/requester-utils/src/BaseResource.ts
@@ -40,7 +40,7 @@ function getDynamicToken(tokenArgument: (() => Promise<string>) | string): Promi
 // Default rate limits per minute
 const DEFAULT_RATE_LIMITS = Object.freeze({
   // Default rate limit
-  '*': 3000,
+  '**': 3000,
 
   // Import/Export
   'projects/import': 6,
--- a/packages/requester-utils/src/BaseResource.ts
+++ b/packages/requester-utils/src/BaseResource.ts
@@ -37,6 +37,7 @@ function getDynamicToken(tokenArgument: (() => Promise<string>) | string): Promi
   return tokenArgument instanceof Function ? tokenArgument() : Promise.resolve(tokenArgument);
 }
 
+// Default rate limits per minute
 const DEFAULT_RATE_LIMITS = Object.freeze({
   // Default rate limit
   '*': 30,
--- a/packages/requester-utils/src/BaseResource.ts
+++ b/packages/requester-utils/src/BaseResource.ts
@@ -1,4 +1,4 @@
-import { RequesterType, ResourceOptions } from './RequesterUtils';
+import { RequesterType, ResourceOptions, RateLimitOptions } from './RequesterUtils';
 
 export interface RootResourceOptions<C> {
   // TODO: Not actually optional - Need to fix wrapper typing in requestUtils.ts:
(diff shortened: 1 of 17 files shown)


packages/requester-utils/src/BaseResource.ts
```diff
@@ -36,6 +37,47 @@ function getDynamicToken(tokenArgument: (() => Promise<string>) | string): Promi
   return tokenArgument instanceof Function ? tokenArgument() : Promise.resolve(tokenArgument);
 }
 
+const DEFAULT_RATE_LIMITS = Object.freeze({
+  // Default rate limit
```
trevor-vaughan, line 41: I know I could look it up, but it would be great to know the time arguments on these as part of the comment. Is it per second?

packages/requester-utils/src/RequesterUtils.ts
```diff
@@ -149,12 +171,13 @@ export function createRequesterFn(
...
   return (serviceOptions) => {
     const requester: RequesterType = {} as RequesterType;
+    const rateLimiters = createRateLimiters(serviceOptions.rateLimits);
 
     methods.forEach((m) => {
       requester[m] = async (endpoint: string, options: Record<string, unknown>) => {
         const requestOptions = await optionsHandler(serviceOptions, { ...options, method: m });
 
-        return requestHandler(endpoint, requestOptions);
+        return requestHandler(endpoint, { ...requestOptions, rateLimiters });
```
trevor-vaughan, line 180: I think that you might want to make this something like: `requestHandler(endpoint, requestOptions, opts)` where `opts` is a named set of additional parameters that can be expanded over time.

```
opts = { rateLimiters: <whatever>, myNewOpt: <something_else> }
```

The first two parameters are set and you run the risk of something from `rateLimiters` overwriting a setting from `requestOptions` in the current implementation.

It's nice when all of the parameters are named (or positional) but that would be a breaking change and probably not worth it at this point.

trevor-vaughan (commented):
Overall code makes sense to me 👍 

Left a couple of nitpick comments that you can address or ignore as you see fit 🎉

packages/requester-utils/src/BaseResource.ts
jdalrymple: I had to change it so its per minute, since i believe thats how the docs describe it.

packages/requester-utils/src/RequesterUtils.ts
jdalrymple: Thought about that but i figured i would just expand the "requestOptions" variable, which is what ive been doing thus far. Normally id include this logic in that optionsHandler function but I didnt want it to reinitialize everytime a request was made.

packages/requester-utils/src/BaseResource.ts
trevor-vaughan: Per *minute*...wow, that's pretty limiting. Definitely need this patch in then!

packages/requester-utils/src/RequesterUtils.ts
trevor-vaughan: Yeah, that's fair. It's all down in the weeds so a later refactor shouldn't mess anyone up if you decide to do it.

packages/requester-utils/src/BaseResource.ts
jdalrymple: I could change the default to 300 maybe? I tried finding to see if there was a default rate limit but couldnt see anything specific in the docs. The rest though are taken directly from the docs

jdalrymple: Though this older doc says more like 1000: https://about.gitlab.com/blog/2020/12/17/automation-check-in-and-rate-limit-changes-on-gitlab-dot-com/

trevor-vaughan: 1000/minute seems like a reasonable generic. We won't really know until it gets thrown at the wall and it's both an easy override and an easy fix.

jdalrymple:
:upside_down_face: need to debug the integration tests before i can release the canary. Hopefully by the end of the week :crossed_fingers:

jdalrymple:
First look at the tests, i must be missing something because i cant seem to narrow down the source. Might be a little more delayed :(.

jdalrymple:
Just gotta debug the browser tests now!

jdalrymple:
Been hitting my head against this all day. Not sure why things are not working. 

I've tried moving the logic to an addScriptTag for both the scripts located in the html file. However, when trying to addScriptTag the gitbeaker script in the form:

```
//gitbeaker.mjs
import * as gitbeaker from '../../../dist/index.mjs'

window.gitbeaker = gitbeaker;
```

It remains undefined. 

If i change the file to just do: 

```
window.test = "TEST"
``` 

The result is correctly test. 

I've tried using a simple script to import but same undefined result. Gonna step back from this for a bit and come fresh another day -_-

trevor-vaughan:
Did you manage to figure this one out? The only thing I can imagine is that the rate limiting additions are somehow intercepting `window` as well.

jdalrymple:
Not yet! I did post a SO question to figure out what im missing, but havent had the time recently to test the suggestions out yet.

trevor-vaughan:
No worries, I'm just really curious as to what it turns out to be.

PR #3374, opened 2023-08-07
Embedding the request options normalization

### LINKED ISSUE #1515: requesterFn is hard to write because of reliance on secret "options handler"
**Description**

Hi jdalrymple! It's been a while. My team is finally moving onto your wonderful package and only ran into one hiccup. Sorry if this is a Feature request; I'm classifying this as a Bug because it seems more like an oversight in implementing the exposure of `requesterFn`.

We were surprised that we had to set our own content-type on JSON requests, and even more surprised that some arguments (such as source/target branches on creating merge requests) seemed to be filling the request body with `camelCase` versions of the parameter names.

The issue appears to be that the default request functions make use of [sanitizing code](https://github.com/jdalrymple/gitbeaker/blob/master/packages/gitbeaker-requester-utils/src/RequesterUtils.ts#L61-L62) that, in the absence of contrary documentation, a programmer implementing requesterFn would reasonably assume had already been taken care of by the library. Indeed, since improper body params are based on internal local variable names in the library, requestFn implementers need to peer deep into the source code just to confirm the exact way in which body parameters will need to be fixed.

**Steps to reproduce**
Implement a requesterFn assuming that the options provided by the library are appropriate for most use cases.

**Expected behaviour**
The library has fully prepared the headers and body before passing it to the custom requesterFn so that making a request in the most expedient way possible will probably work in most cases.

**Actual behaviour**
Object representation of request body is not fully transformed and prepared for transmission, among other issues with headers and possibly form data.

**Possible fixes**
1. Pre-run the defaultOptionsHandler before passing the results to the appropriate requesterFn, or
2. Alternatively, provide detailed documentation about how to implement requesterFn which conspiculously mentions these caveats.

jdalrymple:
closes #1515

### DIFF
--- a/packages/requester-utils/src/BaseResource.ts
+++ b/packages/requester-utils/src/BaseResource.ts
@@ -1,8 +1,8 @@
-import { DefaultResourceOptions, RequesterType } from './RequesterUtils';
+import { RequesterType, ResourceOptions } from './RequesterUtils';
 
 export interface RootResourceOptions<C> {
   // TODO: Not actually optional - Need to fix wrapper typing in requestUtils.ts:
-  requesterFn?: (resourceOptions: DefaultResourceOptions) => RequesterType;
+  requesterFn?: (resourceOptions: ResourceOptions) => RequesterType;
   host?: string;
   prefixUrl?: string;
   rejectUnauthorized?: boolean;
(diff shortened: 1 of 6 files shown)


jdalrymple:
:rocket: PR was released in [`39.10.3`](https://github.com/jdalrymple/gitbeaker/releases/tag/39.10.3) :rocket:

PR #3381, opened 2023-08-17
Fix MergeRequestSchema.closed_at property type

kouak:
This PR fixes `MergeRequestSchema.closed_at` property type which was wrongly typed as a `UserSchema`.

This should be `string | null`, with the timestamp of when the MR was closed.


jdalrymple:
:rocket: PR was released in [`39.11.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/39.11.0) :rocket:

Issue #3384, opened 2023-08-18
Improve/fix `IssueSchema.labels` / `MergeRequestSchema.labels` typing


kouak:
**Description**

`IssueSchema.labels` is typed as `undefined | string[]` which is correct unless `requestOptions.withLabelsDetails` is set to true.

In that case, Gitlab API does not return an array of string for this property, but an array of objects:

```typescript
const issues = await api.issues.all({ projectId: '...', withLabelsDetails: true });
```

```json5
[
  {
    id: 123,
    iid: 12345,
    // ...
    labels: [
      {
        id: 456,
        name: "bug",
        description: null,
        description_html: "",
        text_color: "#FFFFFF",
        color: "#000000",
      },
    ],
    // ...
  },
]
```

Merge requests have the same issue with the `withLabelDetails` request option.

**Proposal**

##### Easy fix: union type

Type `IssueSchema.labels` as an union : `string[] | IssueLabelDetailsSchema[]`.

This would generate typescript errors for current users.

```typescript
const issues = await api.Issues.all({ projectId: 'foo' });
for(const issue of issues) {
  for(const label of labels) {
    if(label.startsWith('foo')) { // Error !
      continue;
    }
    // ...
  }
}
```

##### Harder fix

IMO the ideal fix would be as follow:

```typescript
issues = await api.Issues.all({ projectId: 'foo' });
issues[0].labels; // undefined | string[]

issues = await api.Issues.all({ projectId: 'foo', withLabelDetails: true });
issues[0].labels; // undefined | IssueLabelDetailsSchema[];
```

I think I've got a solution with generics working for this kind of stuff (i.e, tweaking returned entity types based on request options). I'll follow this up with a draft PR for discussion.

**Checklist**
- [x] I have checked that this is not a duplicate issue.
- [x] I have read the documentation.

jdalrymple:
Could also use a function overload? Might be significantly less work haha

kouak:
I'm not too familiar with overloads in typescript, I usually stick with generics. Better the devil you know :)

I believe you need an overload per combination of input parameters, which could be a pain if there are multiple boolean flags ?

Anyway, here's a boiled down implementation with generics : https://tsplay.dev/N7VoDw

jdalrymple:
[...]

PR #3385, opened 2023-08-18
Make Issues.all() generic, return type is based on request options

kouak:
This is a draft PR for #3384


jdalrymple:
Check out the other PR, #3386 I think it accomplishes what youre looking for with a bit less complexity. The polymorph type would be very useful if there were many different options, but since its just one, it might be overkill.

PR #3395, opened 2023-08-30
Add missing scopes and token attribute

Djcd:
Hi,

i saw that some of the scopes i need for the creation of a user impersonation token are missing. So i added those and also checked the other token types and added the missing scopes accordingly.

I also added the 'token' attribute to every token schema because at creation level the token attribute is available (and my ide drives me crazy ;) ), see here:

https://docs.gitlab.com/ee/api/project_access_tokens.html#create-a-project-access-token
https://docs.gitlab.com/ee/api/group_access_tokens.html#create-a-group-access-token
https://docs.gitlab.com/ee/api/users.html#create-an-impersonation-token
https://docs.gitlab.com/ee/api/users.html#create-a-personal-access-token

### DIFF
--- a/packages/core/src/resources/PersonalAccessTokens.ts
+++ b/packages/core/src/resources/PersonalAccessTokens.ts
@@ -18,14 +18,20 @@ export interface PersonalAccessTokenSchema extends Record<string, unknown> {
   last_used_at: string;
   active: boolean;
   expires_at?: string;
+  token?: string;
 }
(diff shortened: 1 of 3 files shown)


jdalrymple:
:rocket: PR was released in [`39.13.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/39.13.0) :rocket:

PR #3401, opened 2023-09-01
Fixing `straight` option type for Repositories.compare

### LINKED ISSUE #3399: `straight` should be `boolean` or `"true" | "false"` instead of `string`
**Description**

Looking at this method, it appears `straight` is `string`

https://github.com/jdalrymple/gitbeaker/blob/2481b154629477d1497be3abfb78b6b36635b8e0/packages/core/src/resources/Repositories.ts#L102-L117

If I make a comparison with this in mind

```js
await fetch(
	"/api/v4/projects/13718909/repository/compare?from=997a91b00cd3974ab4aae8275be85e98c7656875&to=f7a4fc34fe6f63ab2cd6b91cc12ddb83ae8d410a&straight=string",
).then(r => r.json())
// { "error": "straight is invalid" }
```

we see that `straight` doesn't actually accept a `string`. According to the docs, it should be a `boolean`.

https://docs.gitlab.com/ee/api/repositories.html#compare-branches-tags-or-commits

**Possible solutions**

Update the types to match the docs

**Checklist**

jdalrymple:
fixes #3399

### DIFF
--- a/packages/core/src/resources/Repositories.ts
+++ b/packages/core/src/resources/Repositories.ts
@@ -103,7 +103,7 @@ export class Repositories<C extends boolean = false> extends BaseResource<C> {
     projectId: string | number,
     from: string,
     to: string,
-    options?: { fromProjectId?: string | number; straight?: string } & Sudo & ShowExpanded<E>,
+    options?: { fromProjectId?: string | number; straight?: boolean } & Sudo & ShowExpanded<E>,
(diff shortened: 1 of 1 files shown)


jdalrymple:
:rocket: PR was released in [`39.13.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/39.13.0) :rocket:

PR #3540, opened 2024-02-04
Removing native support for rejectUnauthorized - Suggest nodejs env NODE_TLS_REJECT_UNAUTHORIZED=0 instead

### LINKED ISSUE #3505: import() are not supported for service workers
**Description**

- Node.js version: 18
- Gitbeaker version: 39.28.0
- OS & version: MacOS latest

Since the move from @gitbeaker/browser to @gitbeaker/rest it seems like import() has been added to the code. 
The problem is that this is not supported inside browser extensions and service workers at large.

See https://github.com/w3c/webextensions/issues/212 for instance.

Here is one of the offensive line:
```
node_modules/@gitbeaker/rest/dist/index.mjs:9:38
     8 |     if (typeof window !== "object") {
  >  9 |       const { Agent } = await import('https');
  >    |                                      ^^^^^^^
    10 |       options.agent = new Agent({
    11 |         rejectUnauthorized: false
```

**Steps to reproduce**

Use gitbeaker/rest inside a service work.

**Expected behaviour**

Ideally should not use dynamic imports.

**Actual behaviour**

Uses dynamic imports.

**Checklist**

jdalrymple:
fixes #3505 

This removed the required dynamic import and recommends better cert practices by users.


FRANK-AND-FREE:
Hi @jdalrymple , is this going to be merged and released soon?
I trying to use gitbeaker as an API client for our company'S GitLab server, but I seem to run into the problem that the "rejectUnauthorized" setting doesn't work correctly.
If I communicate to the server GitLab API directly via axios http requests (setting recjectUnauthorized': false via their httpsAgent) it works fine, but not via gitbeaker.

jdalrymple:
Hello @FRANK-AND-FREE !  Im still debating adding back this feature. The discussion has been had in #3505 and related issues. It might be best to temporarily use a [work around](https://blog.howtoclicks.com/blog/how-to-set-value-for-node_tls_reject_unauthorized/)  and sort out the certificate issues for your GL instance.

jdalrymple:
@FRANK-AND-FREE Did you have any luck with that work around, or did you end up looking into the cert fixes?

jdalrymple:
:rocket: PR was released in [`39.34.3`](https://github.com/jdalrymple/gitbeaker/releases/tag/39.34.3) :rocket:

FRANK-AND-FREE:
> @FRANK-AND-FREE Did you have any luck with that work around, or did you end up looking into the cert fixes?

Hi @jdalrymple, since I only need to call one or two endpoints, I figured I just keep with my working axios solution. I am using you TypeScript type definitions, though. They still come in very handy. ;-)

Thank you for your work!
Best
Frank

PR #3555, opened 2024-03-08
Upgrade testing gitlab version

jdalrymple:


jdalrymple:
For the curious, here is where the fail is

image

But before this i see messages like:

`/opt/gitlab/embedded/bin/runsvdir-start: line 37: /proc/sys/fs/file-max: Read-only file system `
`cat: /var/opt/gitlab/gitlab-rails/VERSION: No such file or directory `

jdalrymple:
It seems to have resolved itself. I should still update the image version though

PR #3570, opened 2024-04-11
Fixing Keyset Pagination sort option from 'dec' to 'desc'

### LINKED ISSUE #3565: Keyset Pagination not working with sort: 'dec', API seems to expect 'desc' instead
**Description**

- Node.js version: 21.6
- Gitbeaker version: 40.0.2
- OS & version: Windows 11

I tried to fetch Project Jobs using the keyset pagination but kept getting 405s. To verify, I replayed the request manually using ThunderClient and got the same response but the additional info "Keyset pagination is not yet available for this type of request". The GitLab API Documentation says that sort should be "asc" or "desc", not "dec" like it's used in the types of gitbeaker. After changing the value of sort, the request worked fine. Also reproduced with Pipelines Endpoint, however the pipelines endpoint throws a 400 error with "sort does not have a valid value" instead, as its keysets requests are less restricted. Maybe it would be good if the "error" property would be exposed to the caller in general.

**Steps to reproduce**

Perform any request with keyset pagination and sort "dec", try again with "desc". Note that this is currently the only keyset request configuration Gitlab supports. Sample request:
```ts
  const jobs = await gitlab.Jobs.all(123456, {
      pagination: "keyset",
      sort: "dec",
      orderBy: "id"
  });
  ```
**Expected behaviour**

Request should be fine.

**Actual behaviour**

Request throws 405 Method not allowed.

**Possible fixes**

Replace "dec" with "desc".

**Checklist**

jdalrymple:
fixes #3565

### DIFF
--- a/packages/core/src/infrastructure/RequestHelper.ts
+++ b/packages/core/src/infrastructure/RequestHelper.ts
@@ -32,7 +32,7 @@ export type PaginationTypes = 'keyset' | 'offset';
 
 export interface KeysetPaginationRequestOptions {
   orderBy: string;
-  sort: 'asc' | 'dec';
+  sort: 'asc' | 'desc';
 }
 
 export interface OffsetPaginationRequestOptions {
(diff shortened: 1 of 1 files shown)


jdalrymple:
:rocket: PR was released in [`40.0.3`](https://github.com/jdalrymple/gitbeaker/releases/tag/40.0.3) :rocket:

PR #3576, opened 2024-04-24
Extend support for the reformat function & add tests for verification

### LINKED ISSUE #3574: PipelineTriggerTokens endpoint incorrectly handles variables via reformatObjectOptions
**Description**

- Node.js version: 20
- Gitbeaker version: 39.30.0
- OS & version: Ubuntu 20.04

We create new pipelines or trigger downstream pipelines contextually depending on whether we're running locally or in an existing CI pipeline. One of the variables passed to the pipeline via GitBeaker's Pipeline.create and PipelineTriggers.trigger endpoints is a valid JSON string (not necessarily URL safe). If the JSON string contains an `=` it breaks the triggers endpoint specifically because reformatObjectOptions splits on `=` [here](https://github.com/jdalrymple/gitbeaker/blob/2f6335a2279bb9a4191e8732ec0889b8e5b2d603/packages/core/src/infrastructure/Utils.ts#L83) without doing any pre-processing of the string beforehand. In our case the way it breaks is in the CI pipeline the JSON is just abruptly cut off where the `=` would have been.

**Steps to reproduce**

```
await api.PipelineTriggerTokens.trigger(<project_id>, 'main', <job-token>, {variables: {CONFIG_JSON: '{"special-var": "=5"}'}})
```

**Expected behaviour**

variable with JSON is properly encoded, just as it is when using the Pipelines.create endpoint, such that the CI Pipeline sees
```
$ echo "${CONFIG_JSON}"
{"special-var": "=5"}
```

**Actual behaviour**

JSON is split up based on the presence of a `=`.
```
$ echo "${CONFIG_JSON}"
{"special-var": "
```

**Possible fixes**

I'm not sure why `reformatObjectOptions` is needed in the first place, but I do recognize that GitLab's API is different across these two endpoints, and the documentation was sparse on the formatting expected for variables in these different endpoints. However, I think it's reasonable to expect that as long as you format the variables object correctly, the create and trigger endpoints should behave roughly the same and pre-process variables in a consistent manner.

**Checklist**

jdalrymple:
fixes #3574

### DIFF
--- a/packages/core/src/infrastructure/Utils.ts
+++ b/packages/core/src/infrastructure/Utils.ts
@@ -74,13 +74,13 @@ export function reformatObjectOptions(
   obj: Record<string, unknown>,
   prefixKey: string,
   decamelizeValues = false,
-) {
+): Record<string, string> {
   const formatted = decamelizeValues ? decamelizeKeys(obj) : obj;
 
   return QS.stringify({ [prefixKey]: formatted }, { encode: false })
     .split('&')
     .reduce((acc: Record<string, string>, cur: string) => {
-      const [key, val] = cur.split('=');
+      const [key, val] = cur.split(/=(.*)/);
 
       acc[key] = val;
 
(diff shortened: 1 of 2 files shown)


jdalrymple:

PR #3582, opened 2024-05-06
Update visibility typing for Groups and Projects APIs

### LINKED ISSUE #3442: Groups not on par with Projects regarding visibility
**Description**

1) For projects, it is possible to directly pass the "visibility" option in order to by visibility (public/internal/private). It is not possible for groups, although the GitLab API supports it.

2) Side question: why is the `visibility` property of `GroupSchema` and `ProjectSchema` a string and not something like `'public' | 'internal' | 'private'`?

**Possible solutions**
- Add a possibility to directly filter groups by visibility. Thus add `visibility?: boolean` to `AllGroupsOptions`
- Modify the `GroupSchema` and `ProjectSchema` types to narrow the type of the `visibility` property

jdalrymple:
Particularly for the Project and Group Schema, as well as the AllGroupsOptions

fixes #3442

### DIFF
--- a/packages/core/src/resources/Groups.ts
+++ b/packages/core/src/resources/Groups.ts
@@ -40,7 +40,7 @@ export interface SimpleGroupSchema extends CondensedGroupSchema {
 export interface GroupSchema extends SimpleGroupSchema {
   path: string;
   description: string;
-  visibility: string;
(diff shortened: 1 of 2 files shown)


jdalrymple:

PR #3588, opened 2024-05-15
Remove token requirement to support access without authentication

### LINKED ISSUE #173: Remove token-requirement on APIs that work without authentication
Some of the APIs GitLab provides are usable without a user generated token, like the `/users/:user_id/projects` path ([reference](https://docs.gitlab.com/ee/api/projects.html#list-user-projects)):

> Get a list of visible projects owned by the given user. When accessed without authentication, only public projects are returned.

It would be awesome to use those without getting an _token is mandatory_ error!

michael-smt:
Fixes #173


jdalrymple:

PR #3605, opened 2024-06-28
Added created_at attribute

nhollander-alert:
Since GitLab 16.11 the Tags API has supported the `created_at` field for annotated tags[^1]. For lightweight tags, this field is `null`.

Tested with the following script:
```ts
import { Gitlab } from "@gitbeaker/rest"
const gl = new Gitlab({token: "glpat-..."});
gl.Tags.all(1234).then((v) => {
    v.forEach((tag) => {
        console.log(`Tag ${tag.name} created ${tag.created_at}`);
    });
});
```

[^1]: https://gitlab.com/gitlab-org/gitlab/-/issues/451011

### DIFF
--- a/packages/core/src/resources/Tags.ts
+++ b/packages/core/src/resources/Tags.ts
@@ -17,6 +17,7 @@ export interface TagSchema extends Record<string, unknown> {
   target: string;
   message?: string;
   protected: boolean;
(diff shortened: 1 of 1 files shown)


jdalrymple:

PR #3615, opened 2024-08-07
feat: qs -> neoqs

PuruVJ:
# Changes

Replaces `qs` with `neoqs`. https://github.com/puruvj/neoqs

Changes:

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,1 @@
-yarn dlx --no-install lint-staged
+yarn install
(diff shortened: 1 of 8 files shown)


.husky/pre-commit
```diff
@@ -1 +1 @@
-yarn dlx --no-install lint-staged
\ No newline at end of file
+yarn install
```
PuruVJ, line 1: This happened on its own when I ran `yarn` 🤔. Changed it back, but then it kept erroring that there is no `--no-install` option

jdalrymple: Yea i fixed it now

PR #3624, opened 2024-09-09
refactor: :label: declare default type first

ArnaudTA:
Hello in you code the return of `allProjects` is `CondensedProjectSchema[]`.
Apparently for ts if no options are provided it matches the first signature.
So I just interved the two signatures to be closer to the api default behavior

ps: topics can be `null` in your schema but i don't see anything on it in the documentation. Feel free to edit my pr to remove it if you think it is a good idea.


jdalrymple:
:rocket: PR was released in [`40.2.1`](https://github.com/jdalrymple/gitbeaker/releases/tag/40.2.1) :rocket:

PR #3641, opened 2024-10-16
Updating `allowedToCreate` type to an array

### LINKED ISSUE #3629: Protected Tags Create typing is wrong
**Description**

- Node.js version: n/a
- Gitbeaker version: main branch
- Gitbeaker release (cli, rest, core, requester-utils): core
- OS & version: n/a

Protecting tags generates a bad request error as the typing on the function is wrong.

**Steps to reproduce**

```typescript
const gitlabApi = new Gitlab({
  host: 'https://some.gitlab.host',
  token: tokenValue
})

await gitlabApi.ProtectedTags.protect(projectId, "v*", {
  allowedToCreate: {
    groupId: GROUP_ID
  },
  createAccessLevel: AccessLevel.MAINTAINER
})
```
This is accepted by the typing but will throw a GitBeaker exception, since the server will response with a bad request.

**Expected behaviour**

It should create the protected tag.

**Actual behaviour**

The server responses with a HTTP status code 400 bad request.

**Possible fixes**

I can get it to work with this workaround:
```typescript
const gitlabApi = new Gitlab({
  host: 'https://some.gitlab.host',
  token: tokenValue
})

await gitlabApi.ProtectedTags.protect(projectId, "v*", {
  allowedToCreate: [
    {
      groupId: GROUP_ID
    }
  ] as unknown as ProtectedTagAccessLevelEntity,
  createAccessLevel: AccessLevel.MAINTAINER
})
```

To me it seems the error is located on this line:
https://github.com/jdalrymple/gitbeaker/blob/main/packages/core/src/resources/ProtectedTags.ts#L53

Instead of accepting a single entity it should accept an array of these entities:
```typescript
allowedToCreate?: ProtectedTagAccessLevelEntity[];
```

The same will then also be true for this line (`protect` function instead of `create`):
https://github.com/jdalrymple/gitbeaker/blob/main/packages/core/src/resources/ProtectedTags.ts#L79

The GitLab documentation also clearly states this has to be a list of entities:
https://docs.gitlab.com/ee/api/protected_tags.html#protect-repository-tags

Also under 16.11:
https://archives.docs.gitlab.com/16.11/ee/api/protected_tags.html#protect-repository-tags

**Checklist**

jdalrymple:
fixes #3629

### DIFF
--- a/packages/core/src/resources/ProtectedTags.ts
+++ b/packages/core/src/resources/ProtectedTags.ts
@@ -50,7 +50,7 @@ export class ProtectedTags<C extends boolean = false> extends BaseResource<C> {
     tagName: string,
     options?: {
       createAccessLevel?: ProtectedTagAccessLevel;
-      allowedToCreate?: ProtectedTagAccessLevelEntity;
+      allowedToCreate?: ProtectedTagAccessLevelEntity[];
     } & Sudo &
       ShowExpanded<E>,
   ): Promise<GitlabAPIResponse<ProtectedTagSchema, C, E, void>> {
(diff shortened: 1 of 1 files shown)


jdalrymple:
:rocket: PR was released in [`40.6.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/40.6.0) :rocket:

PR #3644, opened 2024-10-20
Update `Users.edit` to use FormData when submitting `avatar` updates

### LINKED ISSUE #3611: Type for uploading a user avatar is incorrect and  requires adding `isForm: true` to work
**Description**

- Node.js version: 18.20.3
- Gitbeaker version: `@gitbeaker/rest@40.1.2`
- OS & version: macOS 14.5

The type for the `avatar` field for creating / editing a user is defined as follows: 

https://github.com/jdalrymple/gitbeaker/blob/e9ca9e97f85e4b5bbc9aaeb240bfefe01569478a/packages/core/src/resources/Users.ts#L159

(an object with a content and filepath field). However, this does not work, a "Bad Request" is returned from gitlab. Non-working example:

```
await api.Users.edit(userId, {
  avatar: {
    content: new Blob([imageBuffer], { type: 'image/png' }),
    filepath: 'avatar.png'
  }
}); // throws GitbeakerRequestError: Bad request
```

Searching through the source code for `isForm: true` seems to indicate that there should be an implementation for Users.ts that takes avatar in the above form and converts it to a tuple adding `isForm: true`. As such the following is a workaround right now:

```
await api.Users.edit(userId, {
  avatar: [new Blob([imageBuffer], { type: 'image/png' }), 'avatar.png'],
  isForm: true
});
```

**Steps to reproduce**

The first snippet above shows a bit of code which reproduces

**Expected behaviour**

We don't get a 400 Bad Request back

**Actual behaviour**

We get a 400 Bad Request back.

**Possible fixes**

Add handling in Users.ts to take avatar in the format indicated by the type into form format.

**Checklist**

jdalrymple:
This should allow the payload to accept the correct Blob types needed for updating the avatar.

fixes #3611

### DIFF
--- a/packages/core/src/resources/Users.ts
+++ b/packages/core/src/resources/Users.ts
@@ -156,7 +156,7 @@ export type AllUsersOptions = {
 export type CreateUserOptions = {
   admin?: boolean;
   auditor?: boolean;
-  avatar?: { content: Blob; filepath?: string };
+  avatar?: { content: Blob; filename?: string };
   bio?: string;
   canCreateGroup?: boolean;
   colorSchemeId?: number;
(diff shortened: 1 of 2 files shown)


jdalrymple:
:rocket: PR was released in [`41.1.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/41.1.0) :rocket:

Issue #3648, opened 2024-10-28
`DeployToken.create` 's API is more limited than underlying GitLab REST API


mpsanchis:
**Description**

The current implementation of `DeployToken.create` allows to pass an arbitrary list of arguments in the options object (in JS), but the type definitions don't. Therefore, when using strict Typescript, we cannot pass options `username` and `expires_at` unless we add something like `@ts-ignore`.

**Expected behaviour**

The `DeployToken.create` API should match the underlying Gitlab REST API: https://docs.gitlab.com/ee/api/deploy_tokens.html#create-a-project-deploy-token

Which exposes the parameters:
* name
* scopes
* expires_at
* username

**Actual behaviour**

The current implementation of the `DeployToken.create` is:
```typescript
  create<E extends boolean = false>(
    name: string,
    scopes: string[],
    {
      projectId,
      groupId,
      ...options
    }: OneOf<{ projectId: string | number; groupId: string | number }> &
      Sudo &
      ShowExpanded<E> = {} as any,
  ): Promise<GitlabAPIResponse<DeployTokenSchema, C, E, void>> {
  ```
  
  Which exposes the parameters:
  * name
  * scopes
  
  and is therefore missing:
  * username
  * expires_at

**Possible fixes**

We could:
a. Simply add them to the function definition as optional (`?`) parameters, or
b. If that is exposing too many parameters, we could create a `DeployTokenCreateOptions` type that contains them all.

I would do (b.), I think it looks cleaner to have something like:
```typescript
create<E extends boolean = false>(
  {projectId, groupId}: OneOf<{ projectId: string | number; groupId: string | number }>,
  options: DeployTokenCreateOptions & Sudo & ShowExpanded<E> = {} as any
): Promise<GitlabAPIResponse<DeployTokenSchema, C, E, void>> {
```

and then define the options that Gitlab's REST API exposes in a:
```typescript
interface DeployTokenCreateOptions {
  name: string,
  scopes: ('read_repository' | 'read_registry' | 'write_registry' | 'read_package_registry' | 'write_package_registry' )[],
  expires_at?: string,
  username?: string,
}
```

**Checklist**

- [X] I have checked that this is not a duplicate issue.
- [X] I have read the documentation.

**Note**

I'd be happy to open a pull request, but I saw in the contribution guide that you'd rather have an issue/discussion first. Let me know what's your preferred approach 😄

jdalrymple:
[...]

PR #3652, opened 2024-10-30
Add options `username` and `expires_at` to DeployTokens.create

mpsanchis:
Fixes https://github.com/jdalrymple/gitbeaker/issues/3648

Some notes:
1. I have added the `DeployTokenCreateOptions` interface as discussed in the issue, and I think it is an understandable API. However, it breaks the previous contract (where `name` and `scopes` were the first two parameters passed). I guess this means a major increase in SemVer.
2. I am not sure how you update versions. I haven't modified the `package.json` because I assume there is some script that does it, but let me know otherwise (it also depends on (1.))
3. The pre-commit hook in husky attempts to run `yarn test`, but there is no `test` script in the `package.json`, so git didn't allow me to commit without `--no-verify`. I skipped the hooks to be able to commit. Am I doing something wrong here?

### DIFF
--- a/packages/core/src/resources/DeployTokens.ts
+++ b/packages/core/src/resources/DeployTokens.ts
@@ -50,13 +50,15 @@ export class DeployTokens<C extends boolean = false> extends BaseResource<C> {
 
   create<E extends boolean = false>(
     name: string,
-    scopes: string[],
+    scopes: DeployTokenScope[],
     {
       projectId,
       groupId,
       ...options
-    }: OneOf<{ projectId: string | number; groupId: string | number }> &
-      Sudo &
+    }: OneOf<{ projectId: string | number; groupId: string | number }> & {
+      expires_at?: string;
+      username?: string;
+    } & Sudo &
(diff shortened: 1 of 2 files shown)


packages/core/src/resources/DeployTokens.ts
```diff
@@ -49,15 +56,8 @@ export class DeployTokens<C extends boolean = false> extends BaseResource<C> {
   }
 
   create<E extends boolean = false>(
-    name: string,
```
jdalrymple, line 52: Since these two are requires, keep them as position arguments

mpsanchis:
@jdalrymple anything else you'd like here?

jdalrymple:
LGTM!

jdalrymple:
:rocket: PR was released in [`41.1.2`](https://github.com/jdalrymple/gitbeaker/releases/tag/41.1.2) :rocket:

mpsanchis:
Thanks Justin!

PR #3654, opened 2024-11-02
Update error JSON handling

### LINKED ISSUE #3646: `GitbeakerRequestError.cause.description` has wrong type
**Description**

- Node.js version: `22.8.0`
- Gitbeaker version: `40.6.0`
- OS & version: Alpine 3.20.2

the `GitbeakerRequestError` object has a property `cause.description` which is defined in [`GitbeakerError.ts:4`](https://github.com/jdalrymple/gitbeaker/blob/77a20d94db8c03c70b372f8bffdb1594c18aa514/packages/requester-utils/src/GitbeakerError.ts#L4) as a string, however at runtime this value contains an object.

**Steps to reproduce**

Run the following code snippet (replacing the token and project ID values with any project you have access to).

```ts
import { GitbeakerRequestError } from "@gitbeaker/requester-utils";
import { Gitlab } from "@gitbeaker/rest";

const gl = new Gitlab({
    token: "glpat-..."
});

try {
    await gl.Pipelines.create(12345, "invalid_ref");
} catch(e) {
    if(e instanceof GitbeakerRequestError) {
        console.log("Description type:", typeof e.cause?.description);
        console.log("Description value:", e.cause?.description);
    }
}

```

**Expected behaviour**

The program should print out
```
Description type: string
Description value: Reference not found
```

**Actual behaviour**

The program will output
```
Description type: object
Description value: { base: [ 'Reference not found' ] }
```

**Possible fixes**

I believe that the issue is caused by the handling of the `description` field in [`Requester.ts`](https://github.com/jdalrymple/gitbeaker/blob/77a20d94db8c03c70b372f8bffdb1594c18aa514/packages/rest/src/Requester.ts#L60). The issue could probably be fixed by checking the type of the `message` field and changing the resolution behavior based on its value.

Gitlab's error message formatting is somewhat inconsistent and not always included in the API documentation so it might be the case that there isn't a guaranteed way to get the core error message, but it might be enough to simply check the type of the `message` field and if it is not a string, return a JSON stringified version of the response object.

```diff
@@ -57,7 +57,9 @@
     if (contentType?.includes('application/json')) {
       const output = JSON.parse(content);
   
-      description = output.message;
+      description = typeof output.message === "string"
+        ? output.message
+        : JSON.stringify(output);
     } else {
       description = content;
     }
```

**Checklist**

jdalrymple:
Ensure the both `error` and `message` key terms are used to populate and raise a more targeted error.

fixes #3646

### DIFF
--- a/global.d.ts
+++ b/global.d.ts
@@ -1,1 +1,22 @@
 import 'jest-extended'
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace jest {
+    interface AsymmetricMatchers {
+      toThrowWith(cb: (error: Error) => void): void;
+    }
+    interface Matchers<R> {
+      toThrowWith(cb: (error: Error) => void): R;
+    }
+  }
+}
+
+declare module "expect" {
+  interface AsymmetricMatchers {
+    toThrowWith(cb: (error: Error) => void): void;
+  }
+  interface Matchers<R> {
+    toThrowWith(cb: (error: Error) => void): R;
+  }
+}
--- a/jest.config.base.mjs
+++ b/jest.config.base.mjs
@@ -1,4 +1,5 @@
 export default {
+  rootDir: '../../',
   testEnvironment: 'node',
   testRegex: 'test\\/.*\\.ts$',
   coverageDirectory: 'coverage',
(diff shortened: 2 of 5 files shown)


jdalrymple:
🙃 somehow managed to hit every typing error under the sun here when trying to add tests for this functionality. Ill try and sort this in a smaller repo and follow up

PR #3666, opened 2024-12-16
Accept username in ProjectMembers#add

### LINKED ISSUE #3665: ProjectMembers#add should accept username if user_id is not provided
**Description**

- Node.js version: 22.11.0
- Gitbeaker version: 41.3.0
- Gitbeaker release (cli, rest, core, requester-utils): REST
- OS & version: Manjaro with Linux kernel 6.12

The method `ProjectMembers#add` should be able to receive the username as an alternative to the user_id since the [API allows it](https://docs.gitlab.com/ee/api/members.html#add-a-member-to-a-group-or-project).

**Steps to reproduce**

- Instantiate Gitlab client from @gitbeaker/rest
- Try to call `client.ProjectMembers.add('repo_id', 'username', AccessLevel.MAINTAINER)`

**Expected behaviour**

It should not cause a type error and should call [this API endpoint](https://docs.gitlab.com/ee/api/members.html#add-a-member-to-a-group-or-project) passing the username as the `username` field of the API endpoint payload

**Actual behaviour**

It causes a type error saying 'username' can't be a string, since the method assumes the second parameter is the user_id (source: https://github.com/jdalrymple/gitbeaker/blob/main/packages/core/src/resources/ProjectMembers.ts#L21), which is a number. Even if the type was right, the implementation still passes this attribute as the `userId` of the API (source: https://github.com/jdalrymple/gitbeaker/blob/main/packages/core/src/templates/ResourceMembers.ts#L67).

**Possible fixes**

I see three possibilities here:

- Make the second argument be of the type `string | number`. If it's a number assume it's the id, and if it's a string assume it's the username. This might be the safest backward-compatible option.
- Make the second argument an object of the type `{ username: string } | { userId: number }` and send some of them conditionally when calling the API.
- Make the user_id and username fields part of the `options` argument instead of being an argument on itself.

**Checklist**

talyssonoc:
Fix #3665


jdalrymple:
Ill switch up the name of the argument to be a bit more descriptive, but looks good otherwise!

jdalrymple:
:rocket: PR was released in [`42.0.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/42.0.0) :rocket:

talyssonoc:
Thank you, @jdalrymple!

PR #3691, opened 2025-03-13
RequestHelper: move maxPages from OffsetPaginationRequestOptions to BasePaginationRequestOptions

lheckemann:
maxPages is accounted for in keyset pagination as well as offset pagination, and passing it is useful when one wants to fetch all items of a given type incrementally.


lheckemann:
Never mind, I misunderstood because it turns out the pipelines API [doesn't support keyset pagination yet](https://gitlab.com/gitlab-org/gitlab/-/issues/431632)

PR #3708, opened 2025-04-14
Preserve slashes in endpoint path segments

### LINKED ISSUE #3707: CommitDiscussions.show fails with 404 due to URL encoding issue in repository/commits path
**Description**

- Node.js version: n/a
- Gitbeaker version: 42.2.0
- Gitbeaker release: @gitbeaker/rest
- OS & version: n/a

I encountered a 404 error when trying to use the `CommitDiscussions.show()` method. After investigating, I found that the URL being constructed has an encoding issue with path segments containing slashes.

```javascript
// When I call:
getGitLabClient().CommitDiscussions.show(projectId, commitId, discussionId);

// The resulting URL is:
// https://gitlab.example.com/api/v4/projects/412/repository%2Fcommits/[commit-id]/discussions/[discussion-id]
//                                                         ^^^^^
// Note the encoded slash in "repository%2Fcommits"
```

However, when manually constructing the URL, it works correctly:

```javascript
gitlabClient.requester.get(`projects/${projectId}/repository/commits/${commitId}/discussions/${discussionId}`);
```

**Steps to reproduce**

1. Create a GitLab instance with the REST API
2. Try to get a commit discussion using `CommitDiscussions.show`:

    ```javascript
    import { Gitlab } from '@gitbeaker/rest';
    
    const gitlabClient = new Gitlab({
      host: 'https://gitlab.example.com',
      token: 'your-token'
    });
    
    const projectId = '123';
    const commitId = 'commit-sha';
    const discussionId = 'discussion-id';
    
    // This fails with 404
    const discussion = await gitlabClient.CommitDiscussions.show(projectId, commitId, discussionId);
    ```

**Expected behaviour**

The CommitDiscussions.show method should successfully retrieve the commit discussion by constructing the correct URL.

**Actual behaviour**

The method fails with a 404 error because the URL is constructed with an encoded slash in "repository%2Fcommits", resulting in an invalid API path.

When debugging the URL construction, I found:

```
CommitDiscussions.resource2Type: repository/commits
```

The issue is that when this string is passed to the `endpoint` template function, it encodes the entire string including the slash, which should actually be preserved as a path separator.

**Possible fixes**

Modify the `endpoint` function in `packages/core/src/infrastructure/Utils.ts` to handle path segments that contain slashes.

**Checklist**

liby:
Fixes #3707

### DIFF
--- a/packages/core/src/infrastructure/Utils.ts
+++ b/packages/core/src/infrastructure/Utils.ts
@@ -41,15 +41,39 @@ export function appendFormFromObject(object: Record<string, OptionValueType>): F
 }
 
 /**
- * Normalize GitLab API endpoint by encoding route parameters.
- * @param strings
- * @param values
+ * A special class to mark path segments that should not be URL-encoded.
+ * Used for paths that already contain special characters like '/' that should be preserved.
+ * For example, in API paths like 'repository/commits', the slash should not be encoded to '%2F'.
  */
-export function endpoint(strings: TemplateStringsArray, ...values: (string | number)[]): string {
-  return values.reduce<string>(
-    (string, value, index) => string + encodeURIComponent(value) + strings[index + 1],
-    strings[0],
-  );
+export class RawPathSegment {
+  public readonly value: string;
+
+  constructor(value: string) {
+    this.value = value;
+  }
+
+  toString(): string {
+    return this.value;
+  }
+}
+
+/**
+ * Normalize GitLab API endpoint by encoding route parameters.
+ * Handles special cases where certain path segments (wrapped in RawPathSegment)
+ * should not be URL-encoded.
+ * @param strings Template string parts
+ * @param values Values to be encoded and inserted between string parts
+ */
+export function endpoint(
+  strings: TemplateStringsArray,
+  ...values: (string | number | RawPathSegment)[]
+): string {
+  return values.reduce<string>((result, value, index) => {
+    const encodedValue =
+      value instanceof RawPathSegment ? value.value : encodeURIComponent(String(value));
+
+    return result + encodedValue + strings[index + 1];
+  }, strings[0]);
 }
 
 /**
--- a/packages/core/src/resources/CommitDiscussions.ts
+++ b/packages/core/src/resources/CommitDiscussions.ts
@@ -1,11 +1,12 @@
 import type { BaseResourceOptions } from '@gitbeaker/requester-utils';
 import { ResourceDiscussions } from '../templates';
-import type {
-  GitlabAPIResponse,
-  PaginationRequestOptions,
-  PaginationTypes,
-  ShowExpanded,
-  Sudo,
+import {
+  type GitlabAPIResponse,
+  type PaginationRequestOptions,
+  type PaginationTypes,
+  RawPathSegment,
+  type ShowExpanded,
+  type Sudo,
 } from '../infrastructure';
 import type {
   DiscussionNotePositionOptions,
@@ -68,6 +69,6 @@ export interface CommitDiscussions<C extends boolean = false> extends ResourceDi
 export class CommitDiscussions<C extends boolean = false> extends ResourceDiscussions<C> {
   constructor(options: BaseResourceOptions<C>) {
     /* istanbul ignore next */
-    super('projects', 'repository/commits', options);
+    super('projects', new RawPathSegment('repository/commits'), options);
   }
 }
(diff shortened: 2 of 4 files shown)


packages/core/src/infrastructure/Utils.ts
```diff
@@ -40,16 +40,40 @@ export function appendFormFromObject(object: Record<string, OptionValueType>): F
...
-    (string, value, index) => string + encodeURIComponent(value) + strings[index + 1],
-    strings[0],
-  );
+export function endpoint(
+  strings: TemplateStringsArray,
+  ...values: (string | number | RawPathSegment)[]
+): string {
+  return values.reduce<string>((result, value, index) => {
+    const encodedValue =
+      value instanceof RawPathSegment ? value.value : encodeURIComponent(String(value));
```
jdalrymple, line 73: Could probably just check for / in the value. I can't think of a scenario off hand where that should cause an issue.

liby: That was actually one of the approaches we [initially explored](https://github.com/jdalrymple/gitbeaker/compare/0382bf629175c00e74a13ef7ee3f1327886551e1..9edf35eb3d8814eb32a6f4595f8c2c483dbe13a2). The challenge we encountered is that the GitLab API seems to have conflicting requirements for slash (`/`) encoding depending on the context of the path segment:

1. **Slashes MUST NOT be encoded**: For certain predefined internal path segments, like `repository/commits` (which is passed internally when constructing URLs for `CommitDiscussions`), the slashes must be preserved literally in the final URL. Encoding it to `repository%2Fcommits` results in a 404. Your suggestion handles this case correctly.

2. **Slashes MUST be encoded**: For user-provided identifiers that can contain slashes, such as project paths (e.g., `group/project`), GitLab requires the slash to be URL-encoded as `%2F`. If we pass `group/project` as a `projectId` variable, the `endpoint` function must produce `.../projects/group%2Fproject/....` If it produces `.../projects/group/project/...`, the API routing fails, typically leading to a 404.

A simple check for `/` that skips encoding would fix case (1) but would break case (2), potentially causing issues for any user whose project or group path contains a slash.

That's why we introduced the `RawPathSegment` wrapper class. It allows us to maintain the default, safe behavior of encoding slashes (correct for case 2) while providing an explicit way to opt-out of encoding for specific, known internal path segments (like `repository/commits` in case 1) where the slashes must be preserved.

This approach avoids ambiguity and ensures both API requirements can be met correctly within the `endpoint` function without interfering with each other.

Let me know what you think!

jdalrymple: Following up - Thats a really good point. I totally forgot about the instances where you pass in "groupname/projectname" as a value. Good catch. 

I was digging around to see if there was any other less complex solutions and double checked the encoding function to see if I was even using the right one. Would `encodeURI` instead of `encodeURIComponent` work better? Since in this case im only encoding uri parts and not the query params etc. This should preserve the /'s without encoding them.

liby: Thanks for digging into this further and suggesting `encodeURI`! It's a good thought, as it does preserve slashes.

I looked into the difference, and while `encodeURI` preserves `/`, it unfortunately runs into the opposite problem compared to `encodeURIComponent`:

1.  **`encodeURIComponent` (Current behavior without `RawPathSegment`):**
    *   Encodes `repository/commits` -> `repository%2Fcommits` (Incorrect - leads to 404 for CommitDiscussions)
    *   Encodes `group/project` -> `group%2Fproject` (Correct - needed for GitLab routing)

2.  **`encodeURI` (Suggested change):**
    *   Encodes `repository/commits` -> `repository/commits` (Correct - fixes CommitDiscussions)
    *   Encodes `group/project` -> `group/project` (Incorrect - breaks GitLab routing for namespaced projects)

The core issue remains that the API requires *different* slash encoding behavior depending on whether the segment is a predefined internal path or a user-provided identifier. `encodeURI` would fix the `repository/commits` case but break the `group/project` case by *not* encoding the slash when it needs to be.

It seems `encodeURIComponent` is generally the safer choice for encoding individual path segments (as it handles spaces, special characters, etc., correctly), but we need a specific mechanism to *prevent* it from encoding the slashes in those known internal path segments like `repository/commits`.

That's where the `RawPathSegment` wrapper comes in – it lets us use the generally safer `encodeURIComponent` by default but explicitly bypass it only for those specific segments where slashes must be preserved literally.

Let me know if that makes sense or if you see another angle!

jdalrymple:
:rocket: PR was released in [`42.4.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/42.4.0) :rocket:

PR #3723, opened 2025-05-04
Updating `iterationId` options typing for the GET /issues endpoint from number -> string

n11sh1:


jdalrymple:
Not sure if this is correct - the [docs](https://docs.gitlab.com/api/issues/#list-issues) specify an int, which in this case would be a number

jdalrymple:
Screenshot 2025-07-12 at 11 57 10 PM

Following up, unless im missing something

PR #3736, opened 2025-06-27
Move maxPages under BasePaginationRequestOptions

dlajarretie:
In the current version, `maxPages` is not declared as a valid pagination option for the `keyset` pagination. However it's implemented and it works well, it's just hidden by the typing.

Capture d’écran 2025-06-27 à 20 10 41

It should be exposed as a valid field for both pagination modes `offset` and `keyset` just like the `perPage` field, which is what this PR does.


jdalrymple:
Still on vacay but ill check this out soon!!

jdalrymple:
:rocket: PR was released in [`43.0.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/43.0.0) :rocket:

PR #3737, opened 2025-07-01
Add support for Markdown Uploads

aidinabedi:
Adding missing Markdown Uploads APIs for both [Projects](https://archives.docs.gitlab.com/17.9/api/project_markdown_uploads/) and [Groups](https://archives.docs.gitlab.com/17.9/api/group_markdown_uploads/).

Also updated README and tests accordingly.

PR #3742, opened 2025-07-18
Add missing function definitions for Markdown Uploads

cody-hutchins:
Adds the functionality to call the other download endpoints provided from Gitlab API for [Projects](https://docs.gitlab.com/17.9/api/project_markdown_uploads/) attachments and [Group](https://docs.gitlab.com/17.9/api/group_markdown_uploads/) attachments.

This is just an addition onto the work @aidinabedi did to add in Markdown Uploads with PR: https://github.com/jdalrymple/gitbeaker/pull/3737.

The way it is implemented utilizes a typescript override on the base ResourceMarkdownUploads classes' download function.


packages/core/src/templates/ResourceMarkdownUploads.ts
```diff
@@ -50,8 +50,33 @@ export class ResourceMarkdownUploads<C extends boolean> extends BaseResource<C>
...
+  download<E extends boolean = false>(
+    resourceId: string | number,
+    secret: string,
+    filename: string,
+    options?: Sudo,
+  ): Promise<GitlabAPIResponse<Blob, C, E, void>>;
+
+  download<E extends boolean = false>(
+    resourceId: string | number,
+    uploadIdOrSecret: string | number,
```
jdalrymple, line 64: why the name uploadIdOrSecret here, but secret in the overrides in groups and projects?

cody-hutchins: Sorry for the delay in responding. I picked that value because in the overrides you will either have an `uploadId` *or* a `secret` but in this function it has to handle both function definitions and be the implementation for both. I figured because both arguments would positionally be in the same place (arg2) in the actual function call that naming it `uploadIdOrSecret` would make it clearer what's happening. What do you think we should call it?

aidinabedi:
If we merge these in and then add the same (secret + filename) overrides for `remove`, we should have all the markdown uploads APIs covered.

packages/core/src/templates/ResourceMarkdownUploads.ts
```diff
@@ -50,15 +50,65 @@ export class ResourceMarkdownUploads<C extends boolean> extends BaseResource<C>
...
+  ): Promise<GitlabAPIResponse<void, C, E, void>>;
+
+  remove<E extends boolean = false>(
+    resourceId: string | number,
+    uploadIdOrSecret: string | number,
+    filename?: any,
+    options?: Sudo,
   ): Promise<GitlabAPIResponse<void, C, E, void>> {
-    return RequestHelper.del()(this, endpoint`${resourceId}/uploads/${uploadId}`, options);
+    if (typeof filename === 'object' || filename === undefined) {
```
jdalrymple, line 101: See comment above

cody-hutchins:
I'm not a regular contributor, is there something I need to do to trigger the ci pipline checks to run correctly?

PR #3749, opened 2025-07-22
fix(core): Rename projectId to groupId in types

ffflorian:
I noticed that `GroupLabels.show` has `projectId` as first parameter (instead of `groupId`, like all other functions of `GroupLabels`.

This does not change any functionality, it is just confusing when using TypeScript 🙂


jdalrymple:
:rocket: PR was released in [`43.4.0`](https://github.com/jdalrymple/gitbeaker/releases/tag/43.4.0) :rocket:

PR #3818, opened 2026-01-17
Fixing security dep issues and removing lerna

jdalrymple:


jdalrymple:
Looks like the problem is deeper than lerna -_-, gonna look into some alternatives

PR #3831, opened 2026-02-01
feat: Add JSDoc Anotations to Jobs API

X-Guardian:
## Summary

This PR adds JSDoc documentation to the `Jobs` API resource class and all its methods, improving IDE intellisense and developer experience when working with the package.

## Examples

image

image

## Notes

The text is copied from the GitLab API documentation.

If you are happy with this, I'll add the same to some of the other APIs.


jdalrymple:
While i think its pretty great to have the JSdocs, it is honestly hard to maintain. Its why i tend to redirect to the Gitlab API docs directly if people want to know more about the particular API's docs :( - I can barely keep up with all the API updates AND the typing haha

PR #3845, opened 2026-05-23
feat: Update MergeRequestDiscussions to use MergeRequestDiscussionSchema for type consistency

moh3n9595:
Changes:

- Adds a new `MergeRequestDiscussionSchema` that extends `DiscussionSchema` with `notes?: MergeRequestDiscussionNoteSchema[]`, so MR-specific note fields (`resolved_by`, `resolved_at`, `position`) are surfaced through the typed API.
- Updates `MergeRequestDiscussions.all`, `create`, `resolve`, and `show` (both interface and class) to return `MergeRequestDiscussionSchema` instead of the generic `DiscussionSchema`.

Per the GitLab MR Discussions API docs, notes returned from MR discussion endpoints include resolution metadata (`resolved_by`, `resolved_at`) and diff `position` data that aren't present on notes from other discussion resources. The library already modeled this via `MergeRequestDiscussionNoteSchema`, but the discussion-returning methods still typed their result as `DiscussionSchema`, so consumers like:

```ts
const discussions = await gitlab.MergeRequestDiscussions.all(projectId, mergeRequestIid);
```

only saw the generic `DiscussionNoteSchema` on `discussions[].notes`. This PR threads the MR-specific schema through so the types match the actual response shape.

### DIFF
--- a/packages/core/src/resources/MergeRequestDiscussions.ts
+++ b/packages/core/src/resources/MergeRequestDiscussions.ts
@@ -25,6 +25,10 @@ export interface MergeRequestDiscussionNoteSchema extends DiscussionNoteSchema {
   position?: DiscussionNotePositionSchema;
 }
 
+export interface MergeRequestDiscussionSchema extends DiscussionSchema {
+  notes?: MergeRequestDiscussionNoteSchema[];
+}
+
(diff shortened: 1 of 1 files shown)


jdalrymple:
Fixed in #3846 ! Good catch!!

PR #3846, opened 2026-06-03
chore: Update API support to match latest Gitlab v19.0

### LINKED ISSUE #3506: Is "forked_from_project" a supported field?
I have a usecase of crawling fork relationships to build out lists, and ideally I'd like to be able to crawl both up and down relationships.

Gitlab's rest documentation states that project schemas will contain parent fork info if one exists (under `forked_from_project`), but I don't see the particular field anywhere in gitbeaker

[Link to the relevant section of the Projects api](https://docs.gitlab.com/ee/api/projects.html#get-single-project:~:text=If%20the%20project%20is%20a,the%20upstream%20project%20by%20default.)

jdalrymple:
fixes https://github.com/jdalrymple/gitbeaker/issues/3506


CMWR421:
👏  nice! wen release 😏 😏 😏

